PR #4, opened 2015-02-04
Rewrite the sequencefile package to be faster and more memory-conscious

colinmarc:
This churns all the sequencefile code and some of the index code, mainly for
the reasons of optimization. In particular, allocating a SectionReader for
each key/value was actually pretty disastrous for memory. The new approach
looks more like a bufio.Scanner, and, internally, reuses a single memory slice.

There's some other miscellaneous cleanup.


index/index.go
```diff
@@ -67,9 +73,10 @@ func (index *Index) BuildIndex() error {
 	wg := sync.WaitGroup{}
 	wg.Add(len(index.files))
 
-	for i, file := range index.files {
-		log.Printf("Indexing %s...\n", file.Path)
-		indexer := newIndexer(file, i, index)
+	for i, _ := range index.files {
```
jbalogh: Doesn't golint say to drop the second bindvar?

index/index.go
```diff
@@ -67,9 +73,10 @@ func (index *Index) BuildIndex() error {
...
+		path := index.files[i].file.Name()
```
jbalogh: ooc, why aren't you using `i, file` in the loop anymore?

colinmarc: because that would copy `file.lock`, which govet complains about.

index/index.go
```diff
@@ -107,11 +114,26 @@ func (index *Index) Get(key string) (*io.SectionReader, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	ok := f.reader.Scan()
+	if !ok {
+		if f.reader.Err() != nil {
+			return nil, f.reader.Err()
+		} else {
+			return nil, io.ErrUnexpectedEOF
```
jbalogh: It feels a little weird that you're synthesizing an error here. Why doesn't the error get returned from `Scan()`?

index/index.go
```diff
@@ -107,11 +114,26 @@ func (index *Index) Get(key string) (*io.SectionReader, error) {
...
+		}
+	}
 
-	record, err := index.files[fileIndex].ReadRecordAtOffset(offset)
-	return record.Value, err
+	return f.reader.Value(), err
```
jbalogh: `err` should be `nil`, right?

colinmarc: ah, good point

index/index.go
colinmarc: I'm imitating the `bufio.Scanner` api, and some others that are similar. `Scan()` returns false if it hits EOF or an error, and then the error is available at `f.reader.Err()`.

In this case, since this is the offset we stored earlier, it's not an expected EOF - there should be a record there.

index/index.go
```diff
@@ -150,53 +174,73 @@ func (index *Index) buildFileList() error {
...
 		i.index.count++
 	}
-}
 
-func serializeIndexEntry(fileIndex int, offset int64) []byte {
-	bytes := make([]byte, 12)
-	binary.BigEndian.PutUint32(bytes, uint32(fileIndex))
-	binary.BigEndian.PutUint64(bytes[4:], uint64(offset))
+	if sf.Err() != nil {
+		i.err = err
```
jbalogh: Should this be `i.err = sf.Err()`?

colinmarc: yes

sequencefile/header.go
```diff
@@ -0,0 +1,145 @@
...
+	Version                   int
+	Compression               compression
+	CompressionCodecClassName string
+	KeyClassName              string
+	ValueClassName            string
+	Metadata                  map[string]string
+	SyncMarker                string
+}
+
+// ReadHeader reads the Header from the input stream, and fill in the Header
```
jbalogh: `s/fill/fills`. You could probably do without the comma.

sequencefile/header.go
```diff
@@ -0,0 +1,145 @@
...
+// struct with the values. This should be called when the reader is positioned
+// at the start of the file or input stream, before any records are read.
+//
+// ReadHeader will also validate that the settings of the sequencefile
+// (version, compression, key/value serialization, etc) are compatible.
+func (r *Reader) ReadHeader() error {
+	magic, err := r.consume(4)
+	if err != nil || string(magic[:3]) != "SEQ" {
+		return fmt.Errorf("Error reading magic number: %s", magic)
```
jbalogh: This is slightly misleading if `err != nil` since the message talks about the magic.

sequencefile/reader.go
```diff
@@ -0,0 +1,177 @@
...
+	err    error
+
+	buf   bytes.Buffer
+	key   []byte
+	value []byte
+}
+
+// New returns a new Reader for sequencefiles, reading input from r. If the
+// io.Reader is positioned at the start of a file, you should immediately call
+// ReadHeader to read through the header.
```
jbalogh: I dislike APIs that require me to do things in a certain sequence. Looking at you `http.Request.ParseForm`.

sequencefile/reader.go
```diff
@@ -0,0 +1,177 @@
...
+	b, err := r.consume(4)
+	if err == io.EOF {
+		return false
+	} else if err != nil {
+		r.close(err)
+		return false
+	}
+
+	// Length -1 means a sync marker (obnoxiously encoded as a cast uint32).
+	totalLength := int(int32(binary.BigEndian.Uint32(b)))
```
jbalogh: Why do you need the `int` and the `int32`?

sequencefile/reader.go
colinmarc: Well, the nice thing is that you can use this to read input streams that don't have headers - or start in the middle of a file, like I do in the indexing code. It also means that `New` doesn't return an error, which I've been trying to get out of the habit of doing.

sequencefile/reader.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+
+		r.key = b[keyStart:keyEnd]
+		r.value = b[valueStart:valueEnd]
+	} else {
+		b, err = r.consume(totalKeyLength)
+		if err != nil {
+			r.close(err)
+			return false
+		}
```
jbalogh: Could you share this with the identical lines at the start of the `if` branch?

jbalogh: Oh nvm `totalLength` vs `totalKeyLength`.

sequencefile/reader.go
```diff
@@ -0,0 +1,177 @@
...
+
+		// To discard the value, first we try seeking. If this isn't a file, then we
+		// just copy to /dev/null.
```
jbalogh: ooc, when are you not reading a file?

sequencefile/reader.go
colinmarc: I think I need to to preserve the fact that hadoop writes `-1` as `int32.MAX_VALUE + 1` (but in an unsigned int). If the go runtime decided ints were 64 bits, then casting directly to `int` would lose that fact.

colinmarc: Please tell me if I'm being crazy, though.

sequencefile/reader.go
colinmarc: Not in sequins. But I happen to know that other people want to use this code for stuffs (at which point I may split it out into a proper library). This is a nice bit of future proofing, and it's nice to just require an `io.Reader` instead of requiring an `io.ReadSeeker` just for this optimization

sequencefile/reader.go
```diff
@@ -0,0 +1,177 @@
...
+
+	return r.scan(readValues)
+}
+
+func (r *Reader) consume(n int) ([]byte, error) {
+	r.buf.Reset()
+	r.buf.Grow(n)
+	b := r.buf.Bytes()[:n]
+
+	_, err := io.ReadFull(r.reader, b)
```
jbalogh: As a naïve reader this looks to me like you're reusing the same buffer all the time. Wouldn't subsequent reads overwrite what you've previously read?

colinmarc: Yup! that's the crux of the memory optimization, actually. It's also documented behavior: `The byte slice will be reused after the next call to Scan.`

PR #6, opened 2015-02-04
Build the index serially, rather than in parallel

colinmarc:
There's questionable performance benefits to running one worker per file in
parallel, since we should be limited by leveldb write speed either way. In
benchmarks, doing it serially seems to be slightly slower, but thrash the disk
way less. Disk thrashing can cause the http server latency to increase, so I
think it's worth it.

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -18,30 +18,22 @@ import (
 var ErrNotFound = errors.New("That key doesn't exist.")
 
 type indexFile struct {
-	lock   sync.Mutex
 	file   *os.File
 	reader *sequencefile.Reader
-}
-
-type indexer struct {
-	index     *Index
-	fileIndex int
-	path      string
-	err       error
 }
 
 type Index struct {
 	Path  string
 	Ready bool
 
-	files []indexFile
+	files     []indexFile
+	readLocks []sync.Mutex
 
 	ldb   *leveldb.DB
 	count int
-
-	readLocks []sync.Mutex
 }
 
+// New creates a new Index instance.
 func New(path string) *Index {
 	index := Index{
 		Path:  path,
@@ -51,6 +43,8 @@ func New(path string) *Index {
 	return &index
 }
 
+// BuildIndex reads each of the files and adds a key -> (file, offset) pair
+// to the master index for every key in every file.
 func (index *Index) BuildIndex() error {
 	err := index.buildFileList()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


index/index.go
```diff
@@ -193,59 +121,98 @@ func (index *Index) addFile(subPath string) error {
...
-func newIndexer(path string, fileIndex int, index *Index) indexer {
-	indexer := indexer{
-		index:     index,
-		fileIndex: fileIndex,
-		path:      path,
+func (index *Index) buildIndexOverFile(fileId int, f indexFile) error {
+	entry := make([]byte, 12)
+
+	for {
+		offset, err := f.file.Seek(0, 1)
```
jbalogh: One of those args is the argument and the other is a constant saying where to seek from, right? Can you use that constant?

index/index.go
```diff
@@ -193,59 +121,98 @@ func (index *Index) addFile(subPath string) error {
...
+		index.ldb.Put(f.reader.Key(), entry, nil)
+		index.count++
+	}
+
+	if err := f.reader.Err(); err != nil {
+		return err
 	}
 
-	return indexer
+	log.Println("Finished indexing", f.file.Name())
```
jbalogh: Should this log go in the calling function? You log `Indexing...` in the caller. If you move it out you can `return f.reader.Err()`.

index/index.go
```diff
@@ -193,59 +121,98 @@ func (index *Index) addFile(subPath string) error {
...
+	return nil
 }
 
-func (i *indexer) start() {
-	file, err := os.Open(i.path)
+// Get returns the value for a given key.
+func (index *Index) Get(key string) ([]byte, error) {
+	if !index.Ready {
+		return nil, errors.New("Index isn't finished being built yet.")
```
jbalogh: What happens while an index is getting rebuilt? Will we ever hit this in normal usage?

index/index.go
```diff
@@ -193,59 +121,98 @@ func (index *Index) addFile(subPath string) error {
...
+	}
+
+	if len(bytes) != 12 {
+		return nil, fmt.Errorf("Invalid value length: %d", len(bytes))
+	}
+
+	fileId, offset := deserializeIndexEntry(bytes)
+	f := index.files[fileId]
+
+	readLock := index.readLocks[fileId]
```
jbalogh: Having a lock on the file seems anti-parallel. Why don't you use multiple file objects?

I guess it would be interesting to benchmark before making performance improvements.

index/index.go
colinmarc: good idea!

index/index.go
```diff
@@ -193,59 +121,98 @@ func (index *Index) addFile(subPath string) error {
...
-		offset, err := file.Seek(0, 1)
-		if err != nil {
-			i.err = err
-			return
+	ok := f.reader.Scan()
+	if !ok {
+		if f.reader.Err() != nil {
+			return nil, f.reader.Err()
+		} else {
+			return nil, io.ErrUnexpectedEOF
```
jbalogh: I've seen this block of code somewhere else.

index/index.go
colinmarc: This is just copy pasta, but to answer the question: no, I don't think so, but I'd rather return that error than panic.

index/index.go
colinmarc: (this is also existing code)

Yeah, as far as I can tell, the read path isn't really the problem. But it might be an interesting optimization to use a `sync.Pool` for each file, or something.

index/index.go
colinmarc: :scream:

colinmarc: yeah as I mentioned, I moved some functions around in this PR. Sorry =(

PR #7, opened 2015-03-03
Add the ability to save and read manifests

colinmarc:
When building an index, save a json file with the expected files and sizes. If
the manifest exists when we start up sequins again, load it and use the
existing index rather than building a brand new index.

r? @jpellerin

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -43,16 +43,97 @@ func New(path string) *Index {
 	return &index
 }
 
-// BuildIndex reads each of the files and adds a key -> (file, offset) pair
-// to the master index for every key in every file.
-func (index *Index) BuildIndex() error {
+// Load reads each of the files and adds a key -> (file, offset) pair to the
+// master index for every key in every file. If a manifest file and index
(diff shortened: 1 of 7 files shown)


index/index.go
```diff
@@ -43,16 +43,97 @@ func New(path string) *Index {
...
+		if baseName != filepath.Base(manifestFile.Name) {
+			return fmt.Errorf("unmatched file: %s", manifestFile.Name)
+		}
+
+		info, err := indexFile.file.Stat()
+		if err != nil {
+			return err
+		}
+
+		if info.Size() != manifestFile.Size {
```
jpellerin: Would it make sense to keep the mtime of the data files in the manifest as well? Is there any chance of a finding file of the same size, but with different data? Would calculating a hash for each file be too expensive to be worthwhile?

colinmarc: These are good questions. I think a hash would be better than mtime, since this is actually designed to be okay with you moving the data directory wholesale. CRC would probably be pretty quick to calculate, I bet.

PR #8, opened 2015-04-08
Fix race condition when upgrading between DB versions

praboud:
Previously, we didn't do any synchronization on the index, so when we upgrade to a new index, the old one just gets stomped, even if there are requests that are currently using that index.

The strategy for fixing this is as follows: we take out an RW mutex when getting a reference to the index, and bump a refcount for that mutex. When updating to a new index version, the upgrader takes out a write lock on the mutex, swaps them, release the mutex, and then waits for the refcount of the mutex to drain to 0 before destroying it. This is reasonably efficient, since we only need mutual exclusion when getting _a reference to the index_, not when actually using it. The logic for this is mostly factored away in side of the newly created IndexMonitor struct.

r? @colinmarc

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -23,21 +23,25 @@ type indexFile struct {
 }
 
 type Index struct {
-	Path  string
-	Ready bool
+	Path    string
+	Ready   bool
+	Version string
 
 	files     []indexFile
 	readLocks []sync.Mutex
 
 	ldb   *leveldb.DB
 	count int
+
+	refcount sync.WaitGroup
 }
 
 // New creates a new Index instance.
-func New(path string) *Index {
+func New(path, version string) *Index {
 	index := Index{
-		Path:  path,
-		Ready: false,
+		Path:    path,
+		Ready:   false,
+		Version: version,
 	}
 
 	return &index
--- a/sequins.go
+++ b/sequins.go
@@ -21,14 +21,13 @@ type sequinsOptions struct {
 }
 
 type sequins struct {
-	options        sequinsOptions
-	backend        backend.Backend
-	index          *index.Index
-	http           *http.Server
-	currentVersion string
-	started        time.Time
-	updated        time.Time
-	reloadLock     sync.Mutex
+	options      sequinsOptions
+	backend      backend.Backend
+	indexMonitor index.IndexReference
+	http         *http.Server
+	started      time.Time
+	updated      time.Time
+	reloadLock   sync.Mutex
 }
 
 type status struct {
@@ -46,7 +45,7 @@ func newSequins(backend backend.Backend, options sequinsOptions) *sequins {
 	}
 }
 
-func (s *sequins) start(address string) error {
+func (s *sequins) init() error {
 	err := s.refresh()
 	if err != nil {
 		return err
@@ -56,7 +55,16 @@ func (s *sequins) start(address string) error {
 	s.started = now
 	s.updated = now
 
-	defer s.index.Close()
+	return nil
+}
+
+func (s *sequins) start(address string) error {
+	// TODO: we may need a more graceful way of shutting down, since this will
+	// cause requests that start processing after this runs to 500
+	// However, this may not be a problem, since you have to shift traffic to
+	// another instance before shutting down anyway, otherwise you'd have downtime
+
+	defer s.indexMonitor.Replace(nil).Close()
 
 	log.Printf("Listening on %s", address)
 	return http.ListenAndServe(address, s)
@@ -82,7 +90,14 @@ func (s *sequins) refresh() error {
 		return err
 	}
 
-	if version != s.currentVersion {
+	// We can use unsafe ref, since closing the index would not affect the version string
+	var currentVersion string
+	currentIndex := s.indexMonitor.UnsafeGet()
+	if currentIndex != nil {
+		currentVersion = currentIndex.Version
+	}
+
+	if version != currentVersion {
 		path := filepath.Join(s.options.LocalPath, version)
 
 		err := os.Mkdir(path, 0700|os.ModeDir)
@@ -101,17 +116,15 @@ func (s *sequins) refresh() error {
 		}
 
 		log.Printf("Preparing version %s at %s", version, path)
-		index := index.New(path)
+		index := index.New(path, version)
 		err = index.Load()
 		if err != nil {
 			return fmt.Errorf("Error while indexing: %s", err)
 		}
 
 		log.Printf("Switching to version %s!", version)
-		oldIndex := s.index
-		s.currentVersion = version
-		s.index = index
 
+		oldIndex := s.indexMonitor.Replace(index)
 		if oldIndex != nil {
 			oldIndex.Close()
 		}
@@ -124,7 +137,11 @@ func (s *sequins) refresh() error {
 
 func (s *sequins) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if r.URL.Path == "/" {
-		count, err := s.index.Count()
+		index := s.indexMonitor.Get()
+		count, err := index.Count()
+		currentVersion := index.Version
+		s.indexMonitor.Release(index)
+
 		if err != nil {
 			log.Fatal(err)
 			w.WriteHeader(http.StatusInternalServerError)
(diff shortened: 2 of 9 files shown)


index/index_monitor.go
```diff
@@ -0,0 +1,46 @@
+package index
+
+// helpers to atomically switch from using one index to another
+
+import (
+	"sync"
+)
+
+type IndexMonitor struct {
```
colinmarc: Can we just put this functionality into `sequins`, or do we need it separate for testing? it seems a bit heavyweight out on its own.

sequins.go
```diff
@@ -56,7 +55,16 @@ func (s *sequins) start(address string) error {
 	s.started = now
 	s.updated = now
 
-	defer s.index.Close()
+	return nil
+}
+
+func (s *sequins) start(address string) error {
+	// TODO: we may need a more graceful way of shutting down, since this will
```
colinmarc: is the issue that we should wait for the http server to shut down? in that case we could just switch in https://github.com/icub3d/graceful/. That can be a separate thing, though.

sequins.go
```diff
@@ -101,17 +116,15 @@ func (s *sequins) refresh() error {
 		}
 
 		log.Printf("Preparing version %s at %s", version, path)
-		index := index.New(path)
+		index := index.New(path, version)
```
colinmarc: I don't see the corresponding change to `Index.New` in this pr, I think?

sequins_test.go
```diff
@@ -18,14 +23,7 @@ func getSequins(t *testing.T, opts sequinsOptions) *sequins {
...
 
-	go func() {
-		err := s.start("localhost:0")
-		if err != nil {
-			t.Fatal(err)
-		}
-	}()
-
-	time.Sleep(100 * time.Millisecond)
+	assert.Nil(t, s.init())
```
colinmarc: you can use `require.NoError()` here, it's nicer.

sequins.go
```diff
@@ -150,7 +168,14 @@ func (s *sequins) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	key := strings.TrimPrefix(r.URL.Path, "/")
-	res, err := s.index.Get(key)
+
+	currentIndex := s.indexMonitor.AcquireRef()
+	if currentIndex == nil {
+		log.Fatal("Tried to make HTTP request before initializing the index") 
+	}
```
colinmarc: Rather than fataling, maybe we could grab the RWMutex on the initial load, so that we know for sure this can't happen?

sequins_test.go
```diff
@@ -63,6 +61,108 @@ func TestSequins(t *testing.T) {
...
 	backend := backend.NewLocalBackend("test_data/0")
 	s := newSequins(backend, sequinsOptions{"test_data/0", false})
-	err := s.start("localhost:0")
+	err := s.init()
 	assert.Error(t, err)
 }
+
+// make sure that reads that occur during an update DTRT
+func TestSequinsThreadsafe(t *testing.T) {
+	assert.NoError(t, os.RemoveAll("test_scratch"))
```
colinmarc: it'd be a little nicer to grab a tmp directory, I think. (the s3 and hdfs tests do, for example).

colinmarc:
left some comments. thanks so much for chasing this down! I like the approach for fixing it.

index/index_monitor.go
praboud: I like the idea of factoring this out into its own thing, since this functionality seems separate from what the sequins "class" is responsible for. That class already seems to be doing too much, so I'm reluctant to add more. I'm open to finding a different home for this, though.

index/index_monitor.go
```diff
@@ -0,0 +1,46 @@
...
+	index *Index
+}
+
+func (im *IndexMonitor) AcquireRef() *Index {
+	im.rwMutex.RLock()
+	index := im.index
+	if index != nil {
+		index.refcount.Add(1)
+	}
+	im.rwMutex.RUnlock()
```
ebroder: `defer`? Not that it's likely to matter, but guaranteed cleanup mechanics >> theoretically non-guaranteed cleanup mechanics.

sequins.go
praboud: Yeah, that's more or less what I'm alluding to. I just didn't want to stuff this into the same PR.

sequins.go
praboud: The only case in which this could happen is if there is no initial data (or a programming error). I almost wonder if we shouldn't error out if the initial `refresh()` doesn't find any data in the directory.

sequins_test.go
praboud: Ah, agreed.

index/index_monitor.go
colinmarc: hrm, I think what I don't like is that the functionality is split over three types - it's hard to reason about.

maybe we could compromise and generalize it? There's no reason it has to be specific to indexes - you could pass it an `interface{}` and a `sync.WaitGroup`, maybe. Or have it take a `ReferenceCounted` interface, which has a `AddReference` and `RemoveReference` method. I'm not seeing any great solutions, though. 

can a map have a pointer key? then you could keep the `WaitGroup` in this type.

sequins.go
colinmarc: If there're no versions to load from, it should error before it gets to this point (and it should error, because that's better than starting up a http server that always 404s, I think).

index/index_monitor.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"sync"
+)
+
+type IndexMonitor struct {
+	rwMutex sync.RWMutex
+	index *Index
+}
+
+func (im *IndexMonitor) AcquireRef() *Index {
```
colinmarc: regardless of what we end up doing here, can you make sure to add comments on all the exported methods?

sequins.go
praboud: Ah, actually, this is not possible at all (unless you're me, and didn't set the test up right at first). I'll get rid of this line, since it's pretty useless.

index/index.go
```diff
@@ -23,21 +23,25 @@ type indexFile struct {
...
 
 	ldb   *leveldb.DB
 	count int
+
+	refcount sync.WaitGroup
 }
 
 // New creates a new Index instance.
-func New(path string) *Index {
+func New(path, version string) *Index {
```
colinmarc: is `Version` used anywhere?

colinmarc: nevermind, I see it

index/index_monitor.go
praboud: Yeah, this is probably best. I think if this code panics, everything blows up anyway, so we don't really care about what state the mux is in. Still, this is probably incrementally easier to reason about.

index/index_monitor.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"sync"
+)
+
+type IndexReference struct {
+	rwMutex sync.RWMutex
+	index *Index
+}
+
+// Atomically acquire a reference to an index, and increment the refcount
```
colinmarc: These should be of the form `Get returns a reference to the index...`. `golint` will helpfully complain about that, if you run it (although I just realized there are some other golint errors I should fix...)

index/index_monitor.go
```diff
@@ -0,0 +1,60 @@
...
+
+// Replace the current index with a new one, wait for the old one to no longer
+// be in use, then return
+func (im *IndexReference) Replace(newIndex *Index) (oldIndex *Index) {
+	im.rwMutex.Lock()
+	oldIndex = im.index
+	im.index = newIndex
+	im.rwMutex.Unlock()
+
+	// it is impossible for new requests to acquire a ref to the old index
```
colinmarc: This comment doesn't seem that useful

sequins.go
```diff
@@ -82,7 +90,14 @@ func (s *sequins) refresh() error {
 		return err
 	}
 
-	if version != s.currentVersion {
+	// we can use unsafe ref, since closing the index would not affect the version string
```
colinmarc: This comment should be capitalized (sorry for being obnoxious about comments)

sequins_test.go
```diff
@@ -63,6 +62,109 @@ func TestSequins(t *testing.T) {
 func TestSequinsNoValidDirectories(t *testing.T) {
 	backend := backend.NewLocalBackend("test_data/0")
 	s := newSequins(backend, sequinsOptions{"test_data/0", false})
-	err := s.start("localhost:0")
+	err := s.init()
 	assert.Error(t, err)
 }
+
+// make sure that reads that occur during an update DTRT
```
colinmarc: like above, this should be `TestSequinsThreadsafe makes sure that...`

colinmarc:
Looks good! Just had some minor comment comments. Can you squash when you're done, and put a description of the problem in the commit message? The PR description should work.

PR #9, opened 2015-04-09
Expose the version in the ETag header

praboud:
Also add this to the status response for completeness.

r? @colinmarc

### DIFF
--- a/sequins.go
+++ b/sequins.go
@@ -35,6 +35,7 @@ type status struct {
 	Started int64  `json:"started"`
 	Updated int64  `json:"updated"`
 	Count   int    `json:"count"`
+	Version string `json:"version"`
 }
 
 func newSequins(backend backend.Backend, options sequinsOptions) *sequins {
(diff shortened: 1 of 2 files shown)


colinmarc:
reading about this a bit more at http://en.wikipedia.org/wiki/HTTP_ETag, this seems like it would be more semantically appropriate were it different for each individual value. Maybe we should just do `X-Sequins-Version` instead. Thoughts?

praboud:
I think the main difference between how ETags ideally work and how this would work is that we would bump the ETag version every day, even if the KVP mapping hasn't changed. I'm not sure this is a problem that we ever intend to solve with sequins.

I'm not convinced that this use of ETag is actually in violation with the ETag spec. There's enough "may" and "should" in the spec that I don't think it's a violation to bump the version when there isn't a difference in the underlying content.

My preference would be to use ETag for this purpose, instead of creating our own custom header, to be able to take advantage of the existing machinery in HTTP clients that uses this header. However, I'm not married to the idea. For my purposes, whether this uses a custom header or not is pretty unimportant.

colinmarc:
Let's do a custom header. I agree that we're not violating the spec, but I don't see much use in reusing an idiom if we're not really matching it. And I'd be a little worried that HTTP clients' existing expectations wouldn't match in subtle ways.

Other than that, this looks good!

PR #10, opened 2015-04-13
Fix bug in cleanup code

praboud:
Defer semantics are hard.
I missed that when you call `s.indexReference.Replace(nil).Close()`, the
`Close()` is deferred, but not the `Replace()`.
This resulted in immediately nilling the index upon starting up the http
server, which causes the server to panic.

r? @colinmarc

### DIFF
--- a/sequins.go
+++ b/sequins.go
@@ -65,7 +65,9 @@ func (s *sequins) start(address string) error {
 	// However, this may not be a problem, since you have to shift traffic to
 	// another instance before shutting down anyway, otherwise you'd have downtime
 
-	defer s.indexReference.Replace(nil).Close()
+	defer func() {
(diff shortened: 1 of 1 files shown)


colinmarc:
Is there a way we can test for this? Not for this specifically, but for the panic we saw.

praboud:
To catch stuff like this, we need to test at a higher level. The extreme version of this is to `exec` the compiled sequins binary from within the tests, wait for HTTP to come up, then make some test requests to to see if things look sane. I'm not convinced this is that bad an idea - this would cover us from problems in `main.go` when setting up the application, including stuff like this. Thoughts?

If we do something like this, I'd prefer to have it done in a separate PR.

colinmarc:
cool - I think that's a good idea.

PR #11, opened 2015-05-01
Add support for gzip encoding

praboud:
We now look at the `Accept-Encoding` header to allow the client to
request gzip encoding.
Currently, the solution only considers the request to determine the
content encoding. In the future, we may want to augment this so sequins
can reply with the identity encoding for small requests, where gzipping
the request is likely to be counter-productive.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ get-deps:
 	go get github.com/crowdmob/goamz/aws
 	go get github.com/stretchr/testify/assert
 	go get github.com/stretchr/testify/require
+	go get github.com/NYTimes/gziphandler
 
 clean:
 	rm -f sequins sequins-*.tar.gz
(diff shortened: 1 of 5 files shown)


praboud:
cc @colinmarc 

I think we should just punt on letting the server have a say in what encoding is used until it becomes a problem. I'm also not sure how this should even work - if the client has said that gzip is preferred over the identity encoding, should the server ignore that, and just serve identity anyway? How much do we want to respect the client's wishes here?

praboud:
Ugh - it looks like `colinmarc/hdfs` points to a defunct code.google.com repo, and is failing the build. @colinmarc, can you fix hdfs so this builds? Also, this is probably a sign that we should have proper vendoring in the sequins repo.

colinmarc:
Huge +1 to supporting gzip. Can we pull in a library to do this instead? https://github.com/NYTimes/gziphandler/blob/master/gzip.go looks pretty plausible.

praboud:
Ah, that looks good. Definitely +1 to not NIH'ing this. I initially didn't see any solutions that didn't have a hilarious approach to (not) parsing the `Accept-Encoding` header, but that looks like it works.

colinmarc:
I fixed gohdfs, so when it tries again it should be fixed.

colinmarc:
This looks good!

PR #13, opened 2015-05-04
Use temp dirs for downloads in progress (fixes #12)

### LINKED ISSUE #12: Errors during download can result in incorrect data
If an individual file in a new version fails to download, the download is stopped -- but the new version's data directory still exists; so on the next reload cycle, that directory will be loaded, even though it contains an incomplete dataset.

jpellerin:
This is not very DRY, but it passes existing tests. Any suggestions for cleaning it up a bit, or adding new tests to exercise the partial download protection? Is this approach to fixing #12 ok, or do you have a different idea of how to address it?

### DIFF
--- a/backend/hdfs_backend.go
+++ b/backend/hdfs_backend.go
@@ -2,9 +2,7 @@ package backend
 
 import (
 	"fmt"
-	"io/ioutil"
 	"log"
-	"os"
 	"path"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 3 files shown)


backend/hdfs_backend.go
```diff
@@ -42,20 +45,34 @@ func (h *HdfsBackend) LatestVersion(checkForSuccess bool) (string, error) {
...
 	versionPath := path.Join(h.path, version)
 	files, err := h.client.ReadDir(versionPath)
 	if err != nil {
 		return err
 	}
 
+	// To avoid loading an incomplete download (#12), download into a temp dir
+	// then rename the temp dir to destPath only if all downloads succeed.
+	baseDir := path.Dir(destPath)
+	workDir, err := ioutil.TempDir(baseDir, fmt.Sprintf("version-%v", version))
```
colinmarc: I think we can push this up into the toplevel type, and pass the directory as `destPath` here. Does that work?

jpellerin: Good idea! I'll give it a try.

jpellerin:
Updated with greater dry-ness.

backend/hdfs_backend.go
```diff
@@ -42,7 +43,7 @@ func (h *HdfsBackend) LatestVersion(checkForSuccess bool) (string, error) {
 	return "", fmt.Errorf("No valid versions at %s", h.displayURL(h.path))
 }
 
-func (h *HdfsBackend) Download(version string, destPath string) error {
+func (h *HdfsBackend) Download(version string, destPath string) (rterr error) {
```
DanielleSucher: ooc, why name the result param here?

sequins.go
```diff
@@ -138,6 +141,31 @@ func (s *sequins) refresh() error {
...
+	// then rename the temp dir to destPath only if all downloads succeed.
+	baseDir := path.Dir(destPath)
+	workDir, err := ioutil.TempDir(baseDir, fmt.Sprintf("version-%v", version))
+	if err != nil {
+		return err
+	}
+	defer func() {
+		// Clean up the temp download dir in the event of a download error
+		if err := os.RemoveAll(workDir); err != nil && !os.IsNotExist(err) {
+			rterr = err
```
DanielleSucher: Why set `rterr` if you never return without specifying the return value?

backend/hdfs_backend.go
jpellerin: Fixed!

The named result param was so that the deferred cleanup func could return
an error if it encountered one -- only way to return something from a defer
-- but that moved to `download` in `sequins.go`, so this is not needed here
any longer.

sequins.go
jpellerin: This is the magical way that go lets you return values from a defer. Since the result var is named, assigning to it changes the return value. :sparkles: It's... not the best thing about the language.

DanielleSucher: :astonished:

DanielleSucher:
lgtm!

jpellerin:
Rebased. Waiting for the built to run and then will merge if it is green.

PR #18, opened 2016-03-11
Add a 'throttle_loads' config property, and use it improve tests

colinmarc:
The functional cluster tests were sometimes failing because the client
doing HTTP gets sometimes wasn't fast enough to catch version changes
happening under the hood. Hooray for sequins being too fast!

I had thought before about allowing users to throttle loads, for
when they have small datasets and are super sensitive about latency.
It turns out this is a nice property to have when we're testing; we
can use it to slow down loads to an observable level.

### DIFF
--- a/blocks/block_store.go
+++ b/blocks/block_store.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"time"
 
 	"github.com/stripe/sequins/sequencefile"
 )
@@ -93,7 +94,7 @@ func NewFromManifest(path string, selectedPartitions map[int]bool) (*BlockStore,
 // AddFile ingests the key/value pairs from the given sequencefile, writing
 // them out to at least one block. If the data is not partitioned cleanly, it
 // will sort it into blocks as it reads.
-func (store *BlockStore) AddFile(reader *sequencefile.Reader) error {
+func (store *BlockStore) AddFile(reader *sequencefile.Reader, throttle time.Duration) error {
 	savedBlocks := make(map[int][]*Block)
 
 	canAssumePartition := true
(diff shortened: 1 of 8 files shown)


blocks/block_store.go
```diff
@@ -101,6 +102,10 @@ func (store *BlockStore) AddFile(reader *sequencefile.Reader) error {
 	assumedFor := 0
 
 	for reader.Scan() {
+		if throttle != 0 {
```
forklady42: tbc if `throttle_loads` is unset in the config, it defaults to a duration of `0` rather than defaulting to `nil`. I assume that's what's happening but didn't see anything in `toml` about defaults nor in this PR.

cluster_test.go
```diff
@@ -66,20 +72,28 @@ func newTestCluster(t *testing.T) *testCluster {
...
+		zkServers:  zkServers,
+		zkCluster:  zkCluster,
+		testClient: testClient,
 	}
 }
 
 func (tc *testCluster) addSequins() {
-	index := len(tc.sequinses)
-	port := 19590 + index // TODO pick random high port
+	port := randomPort()
```
forklady42: anything we can do to confirm this por` is free? relying on it probably being free is the type of thing that likes to come back and bite us

forklady42:
too fast! what a problem to have :speedboat: 

throttling lgtm-- thanks for making a pr!

blocks/block_store.go
colinmarc: Ah, that's just a golang thing. there are no nil ints.

cluster_test.go
colinmarc: I tried a few things, but didn't come up with anything good. In any case the zookeeper test cluster stuff does the same thing, so even if we fixed it here we'd have the same problem there. ¯\_(ツ)_/¯

Since the tests can be rerun easily, this seems not hugely pressing.

blocks/block_store.go
forklady42: oh! good to know

cluster_test.go
forklady42: Fair, I'd mostly be worried about interfering with anything else that's running, but if the tests fail without interrupting other processes, I'm good with it.

PR #19, opened 2016-03-15
Allow nodes to join an existing cluster immediately

colinmarc:
In cases where a node is joining a cluster that already has a version
ready to go, it should be able to start up and immediately start
serving requests to that version (but exclusively proxying them).

This has the happy side-effect of making the version interface a
little simpler, as well.


backend/backend.go
```diff
@@ -13,9 +13,10 @@ type Backend interface {
 	ListDBs() ([]string, error)
 
 	// ListVersions returns a sorted list of valid versions for the given db. If
-	// checkForSuccess is passed, it will only return versions for which there
-	// is a _SUCCESS file present.
-	ListVersions(db string, checkForSuccessFile bool) ([]string, error)
+	// after is possed, only versions greater than it will be returned. If
```
forklady42: typo: possed instead of passed

forklady42:
Glanced over this and it seems reasonable, though there's enough code changes that it's hard to give a close review.

PR #20, opened 2016-03-17
Lock the local store on startup

colinmarc:


forklady42:
travis build is being temperamental, but otherwise lgtm

colinmarc:
Ah, it's not tempermental, I actually 100% broke it - fixed in 0b7464bf8ab30052c46202416152d7a406706c2c

PR #21, opened 2016-03-17
Add some rough expvars for sequins

colinmarc:
This adds a few config properties for tweaking debug output, and adds
some basic sequins-specific expvars, while we're at it - QPS, latency,
and disk usage.


forklady42:
Yay for monitoring!

:+1: once tests are successful

PR #23, opened 2016-03-19
Swap in sparkey for CDB for block storage

colinmarc:
Sparkey is similar to CDB, but has nice features like block
compression and no maximum file size. Unfortunately, building it
is this whole _thing_.


forklady42:
What's our testing plan here? I think we should give this at least a couple of days of dark reads to confirm that sparkey behaves as expected.

colinmarc:
Yeah, exactly that. It's been deployed over the weekend, and dark reads did catch one issue with the go sparkey wrapper I'm using.

PR #24, opened 2016-03-20
Add partition info status JSON, and wrap it in HTML for browsers

colinmarc:
This adds partition info to the status JSON returned by `GET /` and
`GET /dbname/`. It also adds a nice html version, for browsers.


colinmarc:
The status page looks like this: 

Feels like it's missing something...

status.tmpl
```diff
@@ -0,0 +1,97 @@
+<html>
+  <head>
+    <style type="text/css">
```
forklady42: Let's break the css into a separate file, in order to clean this it up a bit. (Makes it easier to focus on just the template.)

colinmarc: It's less than 100 lines =) If we do that, we have to chunk the CSS file into the binary too, and then write an HTTP handler for it...

colinmarc:
Ok, much nicer, I think:

forklady42:
looks good! I like the new layout. this will be sooooo helpful 👍 

(boo flakey test, but it's just flakey)

PR #25, opened 2016-04-06
Increase zookeeper autoreconnect timer to 10s

forklady42:
Increasing the other sleep to 10s too

### DIFF
--- a/zk_watcher.go
+++ b/zk_watcher.go
@@ -137,6 +137,8 @@ func (w *zkWatcher) cancelWatches() {
 	}
 }
 
+const zkReconnectPeriod = 10 * time.Second
+
 // sync runs the main loop. On any errors, it resets the connection.
 func (w *zkWatcher) run() {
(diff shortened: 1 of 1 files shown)


colinmarc:
:+1:

colinmarc:
Actually let's move this into a `const`.

forklady42:
cool, will update

zk_watcher.go
```diff
@@ -137,6 +137,8 @@ func (w *zkWatcher) cancelWatches() {
 	}
 }
 
+const waitPeriod = 10 * time.Second
```
colinmarc: this is a const in the whole package, so it should be named more specifically. perhaps `zkReconnectPeriod` or similar?

forklady42: great point! updated to `zkReconnectPeriod`

colinmarc:
:+1: !!!

PR #26, opened 2016-04-06
If files have changed, remove the new blockstore

forklady42:


colinmarc:
want to add a comment? other than that, :+1:

PR #27, opened 2016-04-15
Swap in gozk for go-zookeeper

colinmarc:
go-zookeeper seems to be giving us trouble. gozk is a bit old, and
perhaps unmaintained, but it's also only ~1000 lines of go; it wraps
the C client. To use it, we have to do this whole vendoring thing
again.

This also makes some cosmetic changes to the logging in zk-watcher,
and inverts the run() loop (since it's a bit cleaner this way).
Finally, it sneaks in some logging of peers, to aid in debugging.

r? @forklady42 

The changes to zk_watcher.go are hidden at the bottom of the diff, because github is dumb.


colinmarc:
There, split the vendoring out into a separate commit so it's a bit easier to review.

colinmarc:
Yessssss got travis and jenkins to build it.

peers.go
```diff
@@ -103,6 +105,8 @@ func (p *peers) updatePeers(addrs []string) {
 		}
 	}
 
+	log.Println("Peers: ", disp)
```
forklady42: This is going to change my life.

zk_watcher.go
```diff
@@ -9,28 +9,23 @@ import (
...
 	"time"
 
-	"github.com/samuel/go-zookeeper/zk"
+	zk "launchpad.net/gozk/zookeeper"
 )
 
 // TODO testable
 
 const coordinationVersion = "v1"
+const zkReconnectPeriod = 1 * time.Second
```
forklady42: :+1: to reducing back to 1 second

forklady42:
Looks good, though switching out the zk client is a pretty key part. What's the game plan for testing before opening the flood gates to prod?

colinmarc:
Good question. I've been doing some tests with an intentionally flaky zookeeper locally; but that's not really the same, obviously. 

At a minimum, we can canary this to a single node. That should be safe against classes of bugs where it would only affect the node with the new client, but that's only half of them.

forklady42:
Hmmm.... Having nodes using different zk clients should be okay as they're both backed by zookeeper, though it does seem a bit ugly. Canarying and then rolling out slowly (during working hours) could work.

I'd feel better spinning a small (three node?) additional cluster and running tests against that.

colinmarc:
Yeah let's hang out and work on that next week.

PR #28, opened 2016-04-16
Make sure we close block files

colinmarc:
This was leaking file descriptors and memory.

r? @forklady42


forklady42:
:ship: :it:

PR #29, opened 2016-04-16
Do our best to return data for misproxied requests

colinmarc:
Under the current behavior, sequins will 501 if asked for a version
of a dataset it doesn't have. This is great from a correctness
standpoint - it means that we can gaurantee a single sequins node
will produce a consistent view, with no version slippage - but it
means that when sharding breaks for any other reason, the results can
be pretty dramatic.

It's important to emphasize availability in this (rare) case. Most
clients don't care if they see version A -> B -> A - they just want
some data. And we return the version with the data, so clients
can independently verify they're getting the right version anyway.

This should give us a strong second line of defense against breakages.

r? @forklady42

### DIFF
--- a/version_mux.go
+++ b/version_mux.go
@@ -54,13 +54,20 @@ func (mux *versionMux) serveKey(w http.ResponseWriter, r *http.Request, key stri
 	if proxyVersion != "" {
 		vs = mux.getVersion(proxyVersion)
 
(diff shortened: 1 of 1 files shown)


forklady42:
Yes to availability!
[selfie-0](https://github.com/thieman/github-selfies/)

PR #30, opened 2016-04-18
ZK Problems

colinmarc:
This downgrades the vendored zookeeper client to 3.4.8 (it was 
from HEAD), and fixes some assorted zookeeper issues.

The biggest change is to increase the timeout. 100ms or 1s is 
way too low of a timeout - I just observed a zookeeper fsync for
1.4s. That may be another issue, but it looks like storm sets 20s
by default, so I think we're in good company with 10s. There's not
really a downside to having a larger timeout, in any case.

This also tweaks some logic; it changes to the reconnect logic such
that it doesn't bother trying to reestablish the session, and it
ensures that we don't try to set watches while we're reconnecting.


zk_watcher.go
```diff
@@ -112,9 +107,6 @@ func (w *zkWatcher) reconnect() error {
 		return err
 	}
 
-	log.Println("Successfully connected to zookeeper!")
```
forklady42: keep this log? we've generally wanted more logging rather than less so far

forklady42:
lgtm 👍

colinmarc:
@forklady42 I added two more tweaks: 2b3b979 and 4def7b2. Want to take another look?

forklady42:
Updates look 😎

PR #32, opened 2016-04-22
Use aws-sdk-go for s3 access

forklady42:
Tests now succeed locally; still need to add expected environment variables to testing environment. Also, should rebase again and squash some of these commits. Plus, there's a bit of refactoring to do, which I may get to this afternoon. Those qualifications aside, this is the bulk of the changes.

r? @colinmarc


main.go
```diff
@@ -109,26 +110,22 @@ func localSetup(localPath string, config sequinsConfig) *sequins {
...
-	if !exists {
-		log.Fatalf("Invalid AWS region: %s", regionName)
-	}
+	// Requiring region name in config for now
+	// if regionName == "" {
+	// 	regionName = aws.InstanceRegion()
+	// 	if regionName == "" {
+	// 		log.Fatal("Unspecified S3 region, and no instance region found.")
+	// 	}
+	// }
```
colinmarc: Can we default to the instance region, still? I'm not sure without looking but I think this'll break our deployment on ec2.

s3_sequins_test.go
```diff
@@ -73,7 +70,46 @@ func TestS3Backend(t *testing.T) {
...
+
+	_, err = svc.PutObject(params)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func putS3Blob(svc *s3.S3, bucket, dst string, src []byte) error {
```
colinmarc: If you make this take an `io.ReadSeeker` or whatever `PutObjectInput` wants, then you can reuse it in the above method.

colinmarc:
(I just added a commit vendoring aws-sdk-go)

main.go
forklady42: Ah, good point! I've now updated this code to use aws-sdk-go.

forklady42:
Ok, I've done some tidying and have added the test environment variables to travis. Ready for review.

colinmarc:
One request - can you rebase everything except the vendor commit into one commit? Github won't let me view the total diff, but seeing the other changes commit-by-commit is hard.

colinmarc:
LGTM!!!!!

PR #33, opened 2016-05-05
Correctly pull in EC2 instance credentials

colinmarc:
r? @forklady42

### DIFF
--- a/main.go
+++ b/main.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -111,17 +112,27 @@ func localSetup(localPath string, config sequinsConfig) *sequins {
 }
 
 func s3Setup(bucketName string, path string, config sequinsConfig) *sequins {
+	metadata := ec2metadata.New(session.New())
```
forklady42: Ha, I had just typed a comment suggesting this but github didn't let it go through. then on refresh, this was here. +1 on great minds thinking alike (...this time at least)

forklady42:
looks good. thanks for the fix!

PR #34, opened 2016-05-05
Use the EC2 instance region if available

colinmarc:
This fixes a bug where the EC2 instance region wasn't being correctly
picked up. The problem is that `regionName, err := metadata.Region()`
instantiates a new variable because of how `:=` works with lexical
scoping.


forklady42:
🚢 🇮🇹

PR #36, opened 2016-05-06
Fix zookeeper issues in tests

colinmarc:
This switches the tests to use the same zk library to create a server as it does to connect to it, so that we can remove samuel/go-zookeeper. It also gives the zk server some time to start up, since that seems to be a problem in travis.

It also fixes the caching of hadoop and zookeeper tarballs in travis, so tests should be faster there.

Yay green builds!


forklady42:
Consolidating zk libraries sgtm. That being said, the travis build failed yet again. /o\

colinmarc:
Noooo it was passing earlier. I ran it like 10 times. argh.

colinmarc:
I know there are still issues to track down, but I just rebuilt like 100 times and can't reproduce them. Merging this as a marginal improvement (and I added some extra logging, so we can track down the issue next time).

forklady42:
cool cool

PR #37, opened 2016-05-08
Implement staged proxying

colinmarc:
This is an attempt to try and minimize variance in latency without increasing load, by introducing a technique in proxying that I'm sure has a real name, but which I'm calling staged proxying.

The basic idea is this. When we need to proxy a request, and have N peers which could service that request, we can try concurrent requests if the peer we try first is being particularly slow. We use the following algorithm to dispatch requests:

 - Each interval of `proxy_stage_timeout`, starting immediately, a request is kicked off to one random not-yet-tried peer. All requests after the first run concurrently.
 - If a request finishes successfully to any peer, the result is returned immediately and the others are canceled, as long as the result was not an error.
 - If a request finishes, but resulted in an error, another is kicked off immediately to another random not-yet-tried peer.
 - This process continues until either all peers are being tried, in which case the code just waits for one to finish. If the total `proxy_timeout` is hit at any point, the method returns immediately with an error and cancels any running requests.

That means we should be able to crank `proxy_stage_timeout` (btw, not in love with this name) all the way down to our observed p99 of a few ms, but keep the `proxy_timeout` high for the cases where we're actually in trouble but want to stay available. Which means that even if one node is behaving really badly, the max latency should never jump above twice the p99. 

In the worst case, where everything is slow, I don't believe this will increase load significantly, since we're actually changing the meaning of `proxy_timeout`. With the old behavior and a `proxy_timeout` of 100ms, we'd do three requests over 300ms, and now we'd do three requests over 100ms. So the requests would be less spread out, but it wouldn't actually increasing the actual number of requests made (and unless the cluster is only fubar for 100ms, that shouldn't make a difference anyway). I think. Please tell me if that sounds wrong.

There are also some config changes to support this. Since `proxy_timeout` now has a semantically different meaning, and I moved a few config properties around, this is a backwards incompatible change. So it's a good thing we haven't cut an RC yet. =)

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -1,0 +1,93 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"log"
+	"net/http"
+	"time"
+)
+
+type proxyResponse struct {
+	resp *http.Response
+	err  error
+}
+
+var errProxyTimeout = errors.New("all peers timed out")
+
+// proxyRequest proxies the request, trying each peer that should have the key
+// in turn. The total logical attempt will be capped at the configured proxy
+// timeout, but individual peers will be tried using the following algorithm:
+//   - Each interval of 'proxy_stage_timeout', starting immediately, a request
+//     is kicked off to one random not-yet-tried peer. All requests after
+//     the first run concurrently.
+//   - If a request finishes successfully to any peer, the result is returned
+//     immediately and the others are canceled, as long as the result was not
+//     an error.
+//   - If a request finishes, but resulted in an error, another is kicked off
+//     immediately to another random not-yet-tried peer.
+//   - This process continues until either all peers are being tried, in which
+//     case the code just waits for one to finish. If the total 'proxy_timeout'
+//     is hit at any point, the method returns immediately with an error and
+//     cancels any running requests.
+func (vs *version) proxyRequest(r *http.Request, peers []string) (*http.Response, error) {
+	responses := make(chan proxyResponse, len(peers))
+	cancel := make(chan struct{})
+	defer close(cancel)
+
+	peerIndex := 0
+	outstanding := 0
+	totalTimeout := time.NewTimer(vs.sequins.config.Sharding.ProxyTimeout.Duration)
+	for {
+		stageTimeout := time.NewTimer(vs.sequins.config.Sharding.ProxyStageTimeout.Duration)
+		if peerIndex < len(peers) {
+			url := fmt.Sprintf("http://%s%s?proxy=%s", peers[peerIndex], r.URL.Path, vs.name)
+			go vs.proxyAttempt(url, responses, cancel)
+			peerIndex += 1
+			outstanding += 1
+		} else if outstanding == 0 {
+			return nil, errNoAvailablePeers
+		}
+
+		select {
+		case res := <-responses:
+			if res.err != nil {
+				log.Printf("Error proxying request to peer: %s", res.err)
+				outstanding -= 1
+				continue
+			} else {
+				return res.resp, nil
+			}
+		case <-totalTimeout.C:
+			return nil, errProxyTimeout
+		case <-stageTimeout.C:
+		}
+	}
+
+	return nil, errNoAvailablePeers
+}
+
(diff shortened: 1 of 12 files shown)


jbalogh:
Google calls this [backup requests](http://highscalability.com/blog/2012/6/18/google-on-latency-tolerant-systems-making-a-predictable-whol.html). /shrug

proxy.go
```diff
@@ -0,0 +1,92 @@
...
+			go vs.proxyAttempt(url, responses, cancel)
+			peerIndex += 1
+		}
+
+		select {
+		case res := <-responses:
+			if res.err != nil {
+				log.Printf("Error proxying request to peer: %s", res.err)
+				// If all peers have errored, don't treat it as a timeout.
+				if peerIndex >= len(peers) {
```
jbalogh: Say we have three peers and the last one errors before we get a success response. In this case, won't we return `errNoAvailablePeers` even though the first two were just slow?

```
start req A
start req B
start req C
req C -> err -> return nil, errNoAvailablePeers
req A -> success
req B -> success
```

colinmarc: Ah, you're totally right. willfix.

version.go
```diff
@@ -104,12 +104,16 @@ func (vs *version) advertiseAndWait() bool {
 
 // serveKey is the entrypoint for incoming HTTP requests.
 func (vs *version) serveKey(w http.ResponseWriter, r *http.Request, key string) {
-	res, err := vs.get(key, r)
+	res, err := vs.get(r, key)
 
 	if err == errNoAvailablePeers {
-		// All of our peers failed us. 502.
-		log.Printf("All peers failed to respond for /%s/%s (version %s)", vs.db, key, vs.name)
+		// Something is wrong with sharding. 502
```
forklady42: It could be something wrong with sharding, but we also error with `errNoAvailablePeers` when every peer errors. I wouldn't necessarily blame that on sharding. I don't know if we want to separate out errors for those two cases or change this description.

forklady42:
Ok, I buy that the load won't be increased too much when everything is slow -- the additional requests will just be closer together. (Load will still be increased, but it will be at most by the replication factor, which is generally fairly small.)

Small comment about errors. All in all, lgtm.

version.go
colinmarc: Good point, I'll change it to be more general.

proxy.go
```diff
@@ -0,0 +1,94 @@
...
+	totalTimeout := time.NewTimer(vs.sequins.config.Sharding.ProxyTimeout.Duration)
+	for {
+		stageTimeout := time.NewTimer(vs.sequins.config.Sharding.ProxyStageTimeout.Duration)
+		if peerIndex < len(peers) {
+			url := fmt.Sprintf("http://%s%s?proxy=%s", peers[peerIndex], r.URL.Path, vs.name)
+			go vs.proxyAttempt(url, responses, cancel)
+			peerIndex += 1
+			outstanding += 1
+		} else if outstanding == 0 {
+			return nil, errNoAvailablePeers
```
jbalogh: Should you `close(cancel)` here?

proxy.go
```diff
@@ -0,0 +1,94 @@
...
+			}
+		case <-totalTimeout.C:
+			close(cancel)
+			return nil, errProxyTimeout
+		case <-stageTimeout.C:
+		}
+	}
+
+	return nil, errNoAvailablePeers
+}
```
jbalogh: It would be cool to keep counters of

* when we kick off a backup request
* when a backup request wins
* when all of them fail and we return `errNoAvailablePeers`

proxy.go
colinmarc: mmm yes.

proxy.go
colinmarc: Yeah, that'd be cool, but I think that depends on better statsd/datadog integration. Maybe I can hack something into the existing expvar stuff. The last one, at least, is covered, since that would result in a 502.

proxy.go
colinmarc: I changed it to `defer close(cancel)` up above, which should be cleaner.

PR #38, opened 2016-05-12
Fix a race condition in staged proxying

colinmarc:
We were closing all open requests when returning from `proxyRequest`, but since we were returning the `http.Response` object, we ocassionally hadn't finished reading the body yet. So `ioutil.ReadAll(resp.Body)` sometimes errored with "request canceled."

r? @jbalogh (since you read the last one so thoroughly)

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -3,6 +3,7 @@ package main
 import (
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


jbalogh:
👍 good catch

PR #39, opened 2016-05-20
Add some zkWatcher tests, and fix some minor bugs

colinmarc:
The new tests caught two (minor) bugs in zkWatcher:

 - `break` inside a switch statement just breaks out of the switch,
   so we weren't properly closing watches on shutdown

 - There was a missing `return` inside `hookWatchChildren` which was
   causing the watch to 'leak'; to continue inside the loop even
   when the connection to zookeeper closed. In some very specific
   conditions, this could cause zkWatcher to infinitely reconnect.

 - We don't need to throw away the connection when we get `STATE_CONNECTING`

 - Access to `w.conn` wasn't synchronized, which lead to a possible (but extremely unlikely) data race

### DIFF
--- a/zk_watcher_test.go
+++ b/zk_watcher_test.go
@@ -9,8 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-
 	zk "launchpad.net/gozk/zookeeper"
 )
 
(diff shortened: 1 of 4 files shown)


zk_watcher_test.go
```diff
@@ -49,28 +64,94 @@ func createTestZk(t *testing.T) testZK {
...
+	go func() {
+		w.createEphemeral("/foo/bar")
+		time.Sleep(100 * time.Millisecond)
+		tzk.restart()
+		w.createEphemeral("/foo/baz")
+	}()
+
+	expectWatchUpdate(t, nil, updates, "the list of children should be updated to be empty first")
+	expectWatchUpdate(t, []string{"bar"}, updates, "the list of children should be updated with the new node")
+	expectWatchUpdate(t, []string{"bar", "baz"}, updates, "the list of children should be updated to be empty again")
```
forklady42: > "the list of children should be updated to be empty again"

This one should be "the list of children should be updated to have two nodes"

colinmarc: Good catch!

forklady42:
👍

colinmarc:
(I just changed the commit message on that last commit)

PR #40, opened 2016-05-20
Add a response header 'X-Sequins-Proxied-To'

scottjab:
# What does this PR do?
Exposes which server handled the request to the client, though a response header `X-Sequins-Proxied-To` 

# Why?
To increase visibility on how many requests are proxied though sequins.  With this data we can chart if requests become slow though proxying or not. 

r @colinmarc

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -9,6 +9,8 @@ import (
 	"time"
 )
 
+const proxyHeader = "X-Sequins-Proxied-To"
+
 type proxyResponse struct {
 	resp *http.Response
 	err  error
@@ -31,7 +33,7 @@ var errProxyTimeout = errors.New("all peers timed out")
 //     case the code just waits for one to finish. If the total 'proxy_timeout'
 //     is hit at any point, the method returns immediately with an error and
 //     cancels any running requests.
-func (vs *version) proxy(r *http.Request, peers []string) ([]byte, error) {
+func (vs *version) proxy(w http.ResponseWriter, r *http.Request, peers []string) ([]byte, error) {
 	responses := make(chan proxyResponse, len(peers))
 	cancel := make(chan struct{})
 	defer close(cancel)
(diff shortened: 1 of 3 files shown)


proxy.go
```diff
@@ -42,12 +44,15 @@ func (vs *version) proxy(r *http.Request, peers []string) ([]byte, error) {
 	for {
 		stageTimeout := time.NewTimer(vs.sequins.config.Sharding.ProxyStageTimeout.Duration)
 		if peerIndex < len(peers) {
-			url := fmt.Sprintf("http://%s%s?proxy=%s", peers[peerIndex], r.URL.Path, vs.name)
-			go vs.proxyAttempt(url, responses, cancel)
+			host := peers[peerIndex]
+			// Should we hit dns?
+			hostName := strings.Split(host, ":")[0]
```
colinmarc: there's a `net.SplitHostPort` you can use here. But I'd actually just include the port in the header, since that's the way we refer to the peer everywhere else.

proxy.go
```diff
@@ -57,7 +61,8 @@ func (vs *version) proxy(r *http.Request, peers []string) ([]byte, error) {
 				outstanding -= 1
 				continue
 			} else {
-				return readResponse(res)
+				response, err := readResponse(res)
```
colinmarc: This is unnecessary

proxy.go
```diff
@@ -42,7 +44,9 @@ func (vs *version) proxy(r *http.Request, peers []string) ([]byte, error) {
 	for {
 		stageTimeout := time.NewTimer(vs.sequins.config.Sharding.ProxyStageTimeout.Duration)
 		if peerIndex < len(peers) {
-			url := fmt.Sprintf("http://%s%s?proxy=%s", peers[peerIndex], r.URL.Path, vs.name)
+			host := peers[peerIndex]
+			w.Header().Add(proxyHeader, host)
```
colinmarc: Maybe add a comment here, to explain what's happening?

proxy_test.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"net"
```
colinmarc: This should go up with the other stdlib imports. Also, `gofmt` and `goimports` should fix it for you.

colinmarc:
Can you squash into one commit? Otherwise, LGTM.

scottjab:
Can do!

PR #41, opened 2016-05-24
Check length of response

forklady42:
@colinmarc


forklady42:
verbal approval from @colinmarc

PR #42, opened 2016-05-24
Check length of response - take 2

forklady42:
@colinmarc


colinmarc:
I think you want `&& len(resp.CommonPrefixes) > 0`

forklady42:
Bah, good call. This is the opposite of the other one. My fingers were too fast.

colinmarc:
shipit

PR #43, opened 2016-05-26
Use ListObjectsPages to properly page keys in S3

colinmarc:
We were checking CommonPrefixes, not Contents, to get the marker for
the next page in listFiles, which meant we were throwing away files
`> 999`. The cleanest way to do this is just to use ListObjectsPages.

Unfortunately, we can't also use ListObjectsPages for listDirs,
because we do a nested request thingy there.

r? @forklady42


forklady42:
lgtm

PR #44, opened 2016-05-27
Fix a race condition when closing a watch

colinmarc:
If the watch was blocked on sending updates to the client process,
then it would panic trying to send on the closed `updates` chan. We can
make sure we catch `cancel`, too, with a `select`.

r? @scottjab


scottjab:
Build is failing, but I don't think on your code? 
otherwise LGTM

PR #46, opened 2016-05-29
Manually dedupe entries when listing files in s3 

colinmarc:
It's not clear why this is necessary, but we've been seeing duplicate
entries fairly often.

We should wait and see if #43 magically fixed this, though.

### DIFF
--- a/backend/s3_backend.go
+++ b/backend/s3_backend.go
@@ -117,7 +117,9 @@ func (s *S3Backend) listDirs(dir, after string) ([]string, error) {
 
 func (s *S3Backend) ListFiles(db, version string) ([]string, error) {
 	versionPrefix := path.Join(s.path, db, version)
-	res := make([]string, 0)
+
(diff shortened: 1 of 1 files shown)


colinmarc:
Nope, #43 did not magically fix this.

backend/s3_backend.go
```diff
@@ -142,8 +144,13 @@ func (s *S3Backend) ListFiles(db, version string) ([]string, error) {
 		return nil, s.s3error(err)
 	}
 
-	sort.Strings(res)
-	return res, nil
+	sorted := make([]string, 0, len(res))
```
forklady42: Kinda amused that this is called sorted even before it's sorted, but it's being created in order to be sorted, so fine.

forklady42:
👍

backend/s3_backend.go
colinmarc: I actually was kinda stuck on what to name it.

PR #47, opened 2016-05-30
Fix a race where we'd leak the zkWatcher reconnect lock

colinmarc:
This was staring me in the face, once I actually went looking for it.
If we `return`ed anywhere between `w.Lock()` and `w.Unlock()` in the
`reconnect` method, we'd leak the lock, and next time around be
unable to reconnect.

r? @scottjab


scottjab:
👍

PR #48, opened 2016-07-25
Use github.com/colinmarc/sequencefile instead of the internal package

colinmarc:
It's newer, and supports compressed sequencefiles.

This diff is kinda hard to read bc `vendor/`, sorry.

r? @kitchen


.gitignore
```diff
@@ -1,5 +1,4 @@
 /sequins
-/sequins-dump
```
kitchen: curious: what does sequins-dump have to do with extracting sequencefile out?

build.go
```diff
@@ -134,12 +136,18 @@ func (vs *version) addFile(bs *blocks.BlockStore, file string) error {
 		return fmt.Errorf("reading %s: %s", disp, err)
 	}
 
-	sf := sequencefile.New(stream)
+	sf := sequencefile.NewReader(bufio.NewReader(stream))
```
kitchen: is this also related to extracting sequencefile?

.gitignore
colinmarc: It was purely a tool to dump sequencefiles; I intend to add a similar tool to the sequencefile library at some point. But it's actually not really useful for sequins, anyway; it used to be that the on-disk storage format was sequencefiles, so you could use it to dump out what sequins had locally. Now that's not the case.

build.go
colinmarc: Yeah, the `sequencefile` in this case changed from the internal library to the external one (in the import statement) and the interface is slightly different.

.gitignore
kitchen: ok! I only ask because I generally try to make unrelated changes as separate commits for compartmentalization purposes. But that's just me! :D

build.go
kitchen: aha, ok. that's what I figured, after I had already made the comment.

kitchen:
LGTM

.gitignore
colinmarc: I generally do too, but I didn't in this case because the intermediate state wouldn't have compiled. (It would've been possible to make that happen, but it didn't seem worth the effort)

colinmarc:
This is currently baking on QA; I'll roll it out in a day or two.

.gitignore
kitchen: aha! totally fine, then :)

PR #49, opened 2016-07-26
Close the block store if we fail to build it

colinmarc:
In the case that we hit an error while adding files, this could leak
file descriptors; We'd close any flushed blocks, but not the ones
open for writing.

r? @kitchen

### DIFF
--- a/build.go
+++ b/build.go
@@ -117,6 +117,8 @@ func (vs *version) createStore(files []string, partitions map[int]bool) (*blocks
 	for _, file := range files {
 		err := vs.addFile(blockStore, file)
 		if err != nil {
+			blockStore.Close()
+			blockStore.Delete()
 			return nil, err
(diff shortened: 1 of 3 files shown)


build.go
```diff
@@ -117,6 +117,8 @@ func (vs *version) createStore(files []string, partitions map[int]bool) (*blocks
 	for _, file := range files {
 		err := vs.addFile(blockStore, file)
 		if err != nil {
+			blockStore.Close()
```
colinmarc: Note that we do this [above](https://github.com/stripe/sequins/blob/a5b3415067a32f710da7f1bb513a95a24ac3ca8a/build.go#L53-L54), if validation of the filesets fails; but not on an actual failure.

kitchen:
seems reasonable!

:shipit:

PR #50, opened 2016-08-01
Support hadoop.io.Text and bare values in addition to hadoop.io.BytesWritable

colinmarc:
Hadoop does this silly key/value serialization thing where it has a serialization class for the key and for the value. In practice, that means that each key and value serialized as [BytesWritable][0] has four bytes wasted at the front for the length of it (wasted because SequenceFile length-prefixes the values too).

I realized while writing the manual that BytesWritable as the only supported key/value serialization was both arbitrary and kinda wasteful. It's [not hard][1] to write "raw" bytes for keys and values, and [Text][2] is another fairly common ["Writable"][3] that uses a single byte to length-prefix (it's actually a weird varint-like thing they call a VInt, but usually it's just one byte).

I figured we may as well support all three. It's easier than being super strict in the docs.

[0]: https://hadoop.apache.org/docs/r2.6.1/api/org/apache/hadoop/io/BytesWritable.html
[1]: https://hadoop.apache.org/docs/current/api/org/apache/hadoop/mapred/SequenceFileAsBinaryOutputFormat.html
[2]: https://hadoop.apache.org/docs/r2.6.1/api/org/apache/hadoop/io/Text.html
[3]: https://hadoop.apache.org/docs/r2.6.1/api/org/apache/hadoop/io/Writable.html

r? @kitchen

### DIFF
--- a/blocks/block_store.go
+++ b/blocks/block_store.go
@@ -2,6 +2,7 @@ package blocks
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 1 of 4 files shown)


kitchen:
seems reasonable!

PR #52, opened 2016-08-15
Bump tylerb/graceful

colinmarc:
This bumps on of our dependencies which has had [a lot of fixes](https://github.com/tylerb/graceful/issues?q=is%3Aissue+is%3Aclosed) in the past few weeks. I have a hunch it may have something to do with our leaked socket issue.

r? @kitchen


kitchen:
:ship:

PR #53, opened 2016-08-17
Rename 'root' to 'source', and add a bunch of path validation

colinmarc:
This renames the `root` configuration option to `source`, as we discussed. It also adds a bunch of validation (and tests for the validation):

  - Check that the source root and local store are absolute paths
  - Check that the former doesn't contain the latter
  - Make sure `file://` urls don't have a "host"
  - Error on schemes other than `file://`, `hdfs://`, or `s3://`

Finally, it adds a bit of code such that relative paths work, but only from the command line.

This is best reviewed commit by commit (since the first one is just s/root/source and is pretty chunderous). r? @kitchen

### DIFF
--- a/main.go
+++ b/main.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"net/url"
+	"path/filepath"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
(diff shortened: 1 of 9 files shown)


main.go
```diff
@@ -21,7 +21,7 @@ var (
 	sequinsVersion string
 
 	bind       = kingpin.Flag("bind", "Address to bind to. Overrides the config option of the same name.").Short('b').PlaceHolder("ADDRESS").String()
-	root       = kingpin.Flag("root", "Where the sequencefiles are. Overrides the config option of the same name.").Short('r').PlaceHolder("URI").String()
+	source     = kingpin.Flag("source", "Where the sequencefiles are. Overrides the config option of the same name.").Short('r').PlaceHolder("URI").String()
```
kitchen: slightly concerned about making this hard flip because this will break existing deployments using `--root` .. maybe deprecate `--root` as a transition mechanism?

colinmarc: Since this is just a development branch, I think it's fine. The migration will be a bit tricky, but not too bad; we just `stripe-puppet` before upping sequins again on each node.

kitchen: ok!

kitchen:
LGTM

PR #54, opened 2016-08-24
Add support for streaming values to clients

colinmarc:
Before this patch, we would buffer values into memory and then dump them out to the client. This is obviously not great for performance.

This PR fixes that, so we stream values either of disk or from peers. In the process, it cleans up some lurking issues with proxying:

 - Headers like `If-Not-Modified` weren't being passed back to peers
 - The `X-Sequins-Proxied-To` header wasn't always set correctly
 - The `X-Sequins-Version` header should be set to the peer's version, even if it's not the one we asked for

Unfortunately, this means we can't easily do http range requests. This PR removes that functionality rather than hack around it, as well as removing functionality for If-Not-Modified (which we could add back later with some more comprehensive thought towards caching and Etags).

### DIFF
--- a/blocks/iterpool.go
+++ b/blocks/iterpool.go
@@ -1,0 +1,46 @@
+package blocks
+
+import (
+	"errors"
+	"fmt"
+	"runtime"
+	"sync"
+
+	"github.com/bsm/go-sparkey"
+)
+
+var errOpenFailed = errors.New("failed to open block iter")
+
+// iterPool is a sync.Pool for *sparkey.HashIter objects. Since each one
+// represents a sparkey buffer in cgo, it's important to reuse them where
+// possible. Since cgo file descriptors don't get closed on GC, we use a
+// finalizer to do that.
+type iterPool struct {
+	sparkeyReader *sparkey.HashReader
+	*sync.Pool
+}
+
+// newIterPool creates a sync.Pool for *sparkey.HashIter objects.
+//
+// The new func returns nil if the sparkey.HashIter couldn't be created.
+func newIterPool(sparkeyReader *sparkey.HashReader) iterPool {
+	return iterPool{
+		sparkeyReader: sparkeyReader,
+		Pool:          new(sync.Pool),
+	}
+}
+
+func (ip iterPool) getIter() (*sparkey.HashIter, error) {
+	iter := ip.Get()
+	if iter == nil {
+		iter, err := ip.sparkeyReader.Iterator()
+		if err != nil {
+			return nil, fmt.Errorf("opening block iter: %s", err)
+		}
+
+		runtime.SetFinalizer(iter, (*sparkey.HashIter).Close)
+		return iter, nil
+	}
+
+	return iter.(*sparkey.HashIter), nil
+}
--- a/blocks/record.go
+++ b/blocks/record.go
@@ -1,0 +1,58 @@
+package blocks
+
+import (
+	"io"
+
+	"github.com/bsm/go-sparkey"
+)
+
+// A Record is one key/value pair loaded from a block.
+type Record struct {
+	ValueLen uint64
+
+	iterPool iterPool
+	iter     *sparkey.HashIter
+	reader   io.Reader
+	closed   bool
+}
+
+func (b *Block) get(key []byte) (*Record, error) {
+	iter, err := b.iterPool.getIter()
+	if err != nil {
+		// In the case of an error, the iter is no longer considered valid.
+		return nil, err
+	}
+
+	if err := iter.Seek(key); err != nil {
+		return nil, err
+	}
+
+	if iter.State() != sparkey.ITERATOR_ACTIVE {
+		// The key doesn't exist, so put the iterator back in the pool.
+		b.iterPool.Put(iter)
+		return nil, nil
+	}
+
+	return &Record{
+		ValueLen: iter.ValueLen(),
+		iterPool: b.iterPool,
+		iter:     iter,
+		reader:   iter.ValueReader(),
+	}, nil
+}
+
+func (r *Record) Read(b []byte) (int, error) {
+	return r.reader.Read(b)
+}
+
+func (r *Record) WriteTo(w io.Writer) (n int64, err error) {
+	return r.reader.(io.WriterTo).WriteTo(w)
+}
+
+func (r *Record) Close() error {
+	if r.iter != nil {
+		r.iterPool.Put(r.iter)
+	}
+
+	return nil
+}
--- a/cluster_test.go
+++ b/cluster_test.go
@@ -562,7 +562,9 @@ func TestClusterNodeWithoutData(t *testing.T) {
 	tc.sequinses[1].config.Test.VersionRemoveTimeout = duration{5 * time.Second}
 	tc.sequinses[2].config.Test.VersionRemoveTimeout = duration{5 * time.Second}
 
-	tc.sequinses[0].expectProgression(down, noVersion, v1, v3)
+	// Because it's behind, it's expected that the first node will flap when it
+	// proxies requests and gets v2 from peers.
+	// tc.sequinses[0].expectProgression(down, noVersion, v1, v3)
 	tc.sequinses[1].expectProgression(down, noVersion, v1, v2, v3)
 	tc.sequinses[2].expectProgression(down, noVersion, v1, v2, v3)
 
--- a/proxy.go
+++ b/proxy.go
@@ -4,9 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"log"
 	"net/http"
+	"net/url"
 	"time"
 )
 
@@ -14,6 +14,7 @@ const proxyHeader = "X-Sequins-Proxied-To"
 
 type proxyResponse struct {
 	resp *http.Response
+	peer string
 	err  error
 }
 
@@ -21,7 +22,8 @@ var errProxyTimeout = errors.New("all peers timed out")
 
 // proxy proxies the request, trying each peer that should have the key
 // in turn. The total logical attempt will be capped at the configured proxy
-// timeout, but individual peers will be tried using the following algorithm:
+// timeout, but individual peers will be tried in the order they are passed in,
+// using the following algorithm:
 //   - Each interval of 'proxy_stage_timeout', starting immediately, a request
 //     is kicked off to one random not-yet-tried peer. All requests after
 //     the first run concurrently.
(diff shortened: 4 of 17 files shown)


colinmarc:
r? @kitchen 

This is best reviewed commit-by-commit.

kitchen:
by "range queries" you mean http range requests, yea? Not "range queries" in the sequins usage of the term (to return a range of keys/values based on a query)

colinmarc:
Sorry, yes, I mean http range requests.

colinmarc:
cc @zenazn - would love some feedback on the go-specific bits (`sync.Pool`, `context`)

blocks/iterpool.go
```diff
@@ -0,0 +1,47 @@
...
+				}
+
+				runtime.SetFinalizer(iter, (*sparkey.HashIter).Close)
+				return iter
+			},
+		},
+	}
+}
+
+func (ip iterPool) getIter() (*sparkey.HashIter, error) {
```
zenazn: I'd just ditch your `New` above and put all the construction logic down here. That way you get to return more useful `error` values straight from the source, and you can spare my eyes 5-deep indentation 🙈

colinmarc: ah, good call.

blocks/record.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/bsm/go-sparkey"
+)
+
+// A Record is one key/value pair loaded from a block.
+type Record struct {
+	ValueLen uint64
+
+	iterPool iterPool
+	iter     *sparkey.HashIter
+	io.Reader
```
zenazn: You lose `sparkey`'s `WriterTo` implementation here

blocks/record.go
```diff
@@ -0,0 +1,46 @@
...
+	if iter.State() == sparkey.ITERATOR_ACTIVE {
+		return &Record{
+			ValueLen: iter.ValueLen(),
+			Reader:   iter.ValueReader(),
+			iterPool: b.iterPool,
+			iter:     iter,
+		}, nil
+	}
+
+	// The key doesn't exist, so put the iterator back in the pool.
```
zenazn: Why don't you put the iterator back in the pool in the other cases?

blocks/record.go
```diff
@@ -0,0 +1,46 @@
...
+	b.iterPool.Put(iter)
+	return nil, nil
+}
+
+func (r *Record) Close() {
```
zenazn: What prevents this from happening twice?

cluster_test.go
```diff
@@ -562,7 +562,9 @@ func TestClusterNodeWithoutData(t *testing.T) {
 	tc.sequinses[1].config.Test.VersionRemoveTimeout = duration{5 * time.Second}
 	tc.sequinses[2].config.Test.VersionRemoveTimeout = duration{5 * time.Second}
 
-	tc.sequinses[0].expectProgression(down, noVersion, v1, v3)
```
zenazn: You misspelled `hobbitses`

proxy.go
```diff
@@ -33,26 +35,22 @@ var errProxyTimeout = errors.New("all peers timed out")
 //     case the code just waits for one to finish. If the total 'proxy_timeout'
 //     is hit at any point, the method returns immediately with an error and
 //     cancels any running requests.
-func (vs *version) proxy(w http.ResponseWriter, r *http.Request, peers []string) ([]byte, error) {
+func (vs *version) proxy(peers []string) (*http.Response, string, error) {
 	responses := make(chan proxyResponse, len(peers))
-	cancel := make(chan struct{})
-	defer close(cancel)
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(vs.sequins.config.Sharding.ProxyTimeout.Duration))
```
zenazn: `r.Context()`

Also would you like to buy a newline

blocks/record.go
colinmarc: I thought I had a [test for it](https://github.com/stripe/sequins/blob/cc49a75f85918e567b73eafc152547b90e719d79/blocks/block_test.go#L57), but now that you say that I see how it's wrong.

I'm not actually sure the right way to do that. Should I add methods that proxy back to the iter?

proxy.go
zenazn: oh nevermind the `http.Request` disappeared

blocks/record.go
zenazn: Yeah—if you know you always have a combination `Reader`/`WriterTo` you can proxy both methods back by hand or similar

proxy.go
```diff
@@ -33,26 +35,22 @@ var errProxyTimeout = errors.New("all peers timed out")
 //     case the code just waits for one to finish. If the total 'proxy_timeout'
 //     is hit at any point, the method returns immediately with an error and
 //     cancels any running requests.
-func (vs *version) proxy(w http.ResponseWriter, r *http.Request, peers []string) ([]byte, error) {
+func (vs *version) proxy(peers []string) (*http.Response, string, error) {
```
zenazn: I might label these return values for documentation purposes. I have no idea what the `string` return is, for instance

blocks/record.go
colinmarc: If there was an error getting it or seeking it, then it's presumed invalid. I'll add more words.

proxy.go
zenazn: This doesn't *feel* like the top of the funnel to me—ideally you have a single `Context` object that connects all the way back to the original request connection (e.g., a `http.Request.Context()`)

colinmarc: Would that let me cancel all this stuff if the original client went away? That does seem useful, but it means lots more threading through.

proxy.go
```diff
@@ -62,51 +60,39 @@ func (vs *version) proxy(w http.ResponseWriter, r *http.Request, peers []string)
...
 		}
 	}
 
-	return nil, errNoAvailablePeers
+	return nil, "", errNoAvailablePeers
 }
 
-func (vs *version) proxyAttempt(url string, res chan proxyResponse, cancel chan struct{}) {
+func (vs *version) proxyAttempt(ctx context.Context, peer string, res chan proxyResponse) {
+	url := fmt.Sprintf("http://%s/%s/%s?proxy=%s", peer, vs.db, vs.name)
```
zenazn: This seems like it's vulnerable to all sorts of fun injection attacks. Maybe construct a `url.URL` object by hand?

serve.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/stripe/sequins/blocks"
+)
+
+// serveKey is the entrypoint for incoming HTTP requests. It looks up the value
+// locally, for, failing that, asks a peer that has it. If the request was
+// already proxied to us, it is not proxied further.
+func (vs *version) serveKey(w http.ResponseWriter, r *http.Request, key string) {
+	// If we don't have any data for this version at all, that's a 404.
+	if vs.numPartitions == 0 {
+		vs.serveNotFound(w)
```
zenazn: do you want to `return` early here?

proxy.go
zenazn: Yup! And I'd wholeheartedly recommend it

serve.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	partition, alternatePartition := blocks.KeyPartition(key, vs.numPartitions)
+	bs := vs.getBlockStore()
+	if bs != nil && vs.hasPartition(partition) || vs.hasPartition(alternatePartition) {
```
zenazn: I'm not sure this is what you mean? https://play.golang.org/p/B4_2xYak6Q

colinmarc: oh dear 😬

serve.go
```diff
@@ -0,0 +1,128 @@
...
+	w.Header().Set("Content-Length", strconv.FormatUint(record.ValueLen, 10))
+	w.Header().Set("Last-Modified", vs.created.UTC().Format(http.TimeFormat))
+	_, err := io.Copy(w, record)
+	if err != nil {
+		// We already wrote a 200 OK, so not much we can do here except log.
+		log.Printf("Error streaming response for /%s/%s (version %s): %s", vs.db, key, vs.name)
+	}
+}
+
+func (vs *version) serveProxied(w http.ResponseWriter, key string, partition, alternatePartition int) {
```
zenazn: Probably not for this PR, but I've seen this `partition` / `alternatePartition` thing a few times now, and it seems like the sort of thing that deserves a (very small) abstraction

serve.go
```diff
@@ -0,0 +1,128 @@
...
+	peers := shuffle(vs.partitions.getPeers(partition))
+	if len(peers) == 0 {
+		log.Printf("No peers available for /%s/%s (version %s)", vs.db, key, vs.name)
+		w.WriteHeader(http.StatusBadGateway)
+		return
+	}
+
+	resp, peer, err := vs.proxy(peers)
+	if err == nil && resp.StatusCode == 404 && alternatePartition != partition {
+		log.Println("Trying alternate partition for pathological key", key)
```
zenazn: When might this happen? What makes a key pathological?

serve.go
```diff
@@ -0,0 +1,128 @@
...
+
+		alternatePeers := shuffle(vs.partitions.getPeers(partition))
```
zenazn: and do you want `alternatePartition` here?

colinmarc: yeah, idk how I managed to mess up copy/paste

proxy.go
```diff
@@ -33,26 +35,22 @@ var errProxyTimeout = errors.New("all peers timed out")
...
 	responses := make(chan proxyResponse, len(peers))
-	cancel := make(chan struct{})
-	defer close(cancel)
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(vs.sequins.config.Sharding.ProxyTimeout.Duration))
+	defer cancel()
 
 	peerIndex := 0
 	outstanding := 0
-	totalTimeout := time.NewTimer(vs.sequins.config.Sharding.ProxyTimeout.Duration)
 	for {
```
zenazn: Curious if you can rephrase this in terms of a `range` over `peers`?

zenazn:
How do you know this is faster / less resource intensive? Is it possible to phrase the improvement as a benchmark, in order to prevent future regressions?

(Also, is it possible to write tests that exercise each of the bugs we've found, just to make sure they don't happen again?)

blocks/record.go
colinmarc: Nothing, but I control the calling code. [`Closer`](https://golang.org/pkg/io/#Closer) says the behavior after the first call is undefined, anyway.

colinmarc: I'm missing the `error` return, though, so I'll fix that.

proxy.go
colinmarc: I tried that first back when I wrote this, but I don't remember why it didn't work well.

serve.go
colinmarc: Read [this comment](https://github.com/stripe/sequins/blob/1.0/blocks/hashcode.go#L31-L78) if you really want to know.

colinmarc: (We've never hit this codepath with our data, but I couldn't ignore the edge case once I thought of it.)

colinmarc:
> How do you know this is faster / less resource intensive? Is it possible to phrase the improvement > as a benchmark, in order to prevent future regressions?

Speed isn't really my goal here, although it's not hard to imagine how this would be faster and less memory-intensive for very large values. The two performance improvements that are secondary goals here - memory usage over an extended period and variability of latency - are really hard to measure with a benchmark.

serve.go
zenazn: jesus fuck

zenazn:
kk; I mostly just wanted to check up on something you wrote in the PR description:

> ...This is obviously not great for performance.
> 
> This PR fixes that...

If that's not a concern, carry on!

colinmarc:
(going to reopen against master)

PR #56, opened 2016-09-08
Prevent a data race when marshalling db statuses

colinmarc:
The 1.7 race detector caught this. We weren't doing a shallow copy of the `Nodes` map, so we we could race mutating it and dumping it out to clients.

I took the opportunity and did a minimal refactor of some janky deep-copy code we had repeated in a few places.

r? @kitchen


kitchen:
LGTM

PR #58, opened 2016-10-05
Allow sequins to start up without data

colinmarc:
This allows us to be a bit more user-friendly in the tutorial; users can start up sequins without any data, and then write data to it second.

r? @kitchen

### DIFF
--- a/main.go
+++ b/main.go
@@ -106,11 +106,9 @@ func main() {
 	}
 
 	// Do a basic test that the backend is valid.
-	dbs, err := s.backend.ListDBs()
+	_, err = s.backend.ListDBs()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -106,11 +106,9 @@ func main() {
 	}
 
 	// Do a basic test that the backend is valid.
-	dbs, err := s.backend.ListDBs()
+	_, err = s.backend.ListDBs()
 	if err != nil {
 		log.Fatalf("Error listing DBs from %s: %s", s.backend.DisplayPath(""), err)
-	} else if len(dbs) == 0 {
```
colinmarc: (here's the actual code change)

kitchen:
LGTM

I think separating the docs changes to a different commit might not be a bad idea, in case someone looks at the history of a line or something!

PR #59, opened 2016-10-05
Revert to using timer-based cancellation for proxied requests

colinmarc:
In c6eb0b, we switched to using context to manage proxying requests to peers and the timeouts therein.

Using context still makes this flow easier, but using a context with a deadline had the unintended side effect that it acts as a deadline for the entire request, not just to connect. Since we return the still-open http.Response from the proxy method, we want to leave it open indefinitely from that point on.

In this way, it was also inadvertently working as a server timeout; we probably want some of those, but that should be addressed separately (and not be exclusive to proxied requests).

r? @zenazn 
cc @kitchen


colinmarc:
Thinking about this, I think this has the problem that it leaks the requests that didn't make it. They'd get closed when the upstream request finishes, but that's not exactly what we want.

proxy.go
```diff
@@ -38,8 +41,8 @@ var errProxyTimeout = errors.New("all peers timed out")
 //     cancels any running requests.
 func (vs *version) proxy(r *http.Request, peers []string) (*http.Response, string, error) {
 	responses := make(chan proxyResponse, len(peers))
-	ctx, _ := context.WithDeadline(r.Context(),
-		time.Now().Add(vs.sequins.config.Sharding.ProxyTimeout.Duration))
+	totalTimeout := time.NewTimer(vs.sequins.config.Sharding.ProxyTimeout.Duration)
+	ctx, cancel := context.WithCancel(r.Context())
```
zenazn: I think you want to `defer cancel()`—otherwise I'm pretty sure you leak resources

colinmarc: maybe I should just move this logic one method up the chain, or something. hrm

colinmarc:
> Any chance I could get you to use a custom http.Transport / http.Client and to set a Timeout on the Transport's Dialer?

I don't think that does what I want, since the whole idea here is to have somewhat dynamic timeouts and backup requests. But maybe I'm misunderstanding what you're suggesting?

colinmarc:
I added some logic to make sure other outstanding requests are canceled when one succeeds, and added a comment to clarify why I'm not `defer cancel()`ing.

kitchen:
Just so I'm reading this correctly:

this is in the proxy fanout, where you fire off a bunch of goroutines to ask ~all of the other nodes for the key. You want to cancel all of them but the one that came back first, because you still need the http.Response object from that one, which you'll cancel after you're done with it (or the context gets GC'd)

Seems legit!

colinmarc:
Yup, that's right!

zenazn:
Apologies—I'm not sure I fully understood the control flow here until my second read through.

One thing that might help is some kind of concise (documented, tested, etc.) abstraction around this "first request past the post" idea. To be honest I'm not really sure what this should look like, and I'm not even sure what level it should act at (`http.Request` and a hijacked `ReadCloser` vs. `context.Context` being two options that come to mind), but I think it would do a lot for understandability.

Here's an untested implementation of the second idea (using `context.Context`) that I'll offer as a starting point for discussion: https://gist.github.com/zenazn/85ec533af61e45448f9d64bc091b70d3

zenazn:
(but
LGTM
I believe these changes, as written, are likely correct)

PR #60, opened 2016-10-18
Build a mechanism for GCing old zookeeper nodes

colinmarc:
Currently, we leak a non-ephemeral znode for every version of every database. This isn't *that* many, but it's kinda annoying and definitely unhygienic. 

However, we can't just go around arbitrarily deleting empty znodes, because that's racy when it comes to other nodes; they could be trying to create an ephemeral node rooted at the node we're deleting. To avoid this, this PR adds some retry logic and creating and watching that, if it needs to, also creates the parents of the nodes.

Finally, this PR fixes one possible deadlock I encountered while poking at this code.

r? @kitchen

### DIFF
--- a/zk_watcher.go
+++ b/zk_watcher.go
@@ -182,15 +182,15 @@ Reconnect:
 				log.Println("Error reconnecting to zookeeper:", err)
 				continue Reconnect
 			}
+
+			// Every time we connect, reset watches and recreate ephemeral nodes.
+			err = w.runHooks()
+			if err != nil {
+				log.Println("Error running zookeeper hooks:", err)
+				continue Reconnect
+			}
 		} else {
 			first = false
-		}
-
-		// Every time we connect, reset watches and recreate ephemeral nodes.
-		err := w.runHooks()
-		if err != nil {
-			log.Println("Error running zookeeper hooks:", err)
-			continue Reconnect
 		}
 
 		select {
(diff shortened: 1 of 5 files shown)


zk_watcher.go
```diff
@@ -382,6 +382,42 @@ func (w *zkWatcher) createAll(node string) error {
...
+	}
+
+	empty := true
+	for _, child := range children {
+		if !w.cleanupTree(path.Join(node, child)) {
+			empty = false
+		}
+	}
+
+	if empty {
```
kitchen: is this not racey with parent node creation?

can we just yolo delete the parent nodes, and if they aren't empty it'll fail and we just move on with life?

colinmarc: Yeah, it's just a minor optimization - it's ok to try and delete and then fail anyway in some race conditions, as you point out. I can remove if you think it's needlessly complicated.

kitchen: I think we'd want to check for that exact situation in this block anyways, so we may as well eliminate the other logic?

colinmarc: Sorry, I don't follow. What do you mean check for that situation? We'll get an error from `w.conn.Delete` below, and that's handled.

kitchen: oh, sorry, I ... didn't read down that far :P

I'm a fan of removing the check since it's racey anyways.

colinmarc: Ok, fixed.

colinmarc:
@kitchen PTAL?

kitchen:
lgtm

PR #62, opened 2016-10-27
Refactor upgrade and build flows

colinmarc:
In service of trying to fix a few lurking bugs in the upgrade flows, I ended up stepping back and doing a somewhat more comprehensive refactor of the surrounding code. This PR altogether is pretty invasive, but each commit stands individually as a smaller refactor, so it hopefully shouldn't be too bad to review. We also fortunately have [pretty good functional tests](https://github.com/stripe/sequins/blob/master/cluster_test.go) for this stuff.

The fixes contained within, in ascending order of importance:
- The status webpage sometimes leaked versions, showing them as errored even after they'd been cleared.
- During startup, builds weren't respecting max_parallel_loads, which meant we could thrash the disk if a node joined a full cluster with nothing cached locally.
- Versions that were still building, but complete on peers, weren't switched to until they finished building. That's now decoupled; when a version is ready to serve, it's switched to.

r? @kitchen
As I mentioned above, this should be reviewed commit-by-commit.

### DIFF
--- a/blocks/block.go
+++ b/blocks/block.go
@@ -29,7 +29,7 @@ type Block struct {
 	sync.RWMutex
 }
 
-func loadBlock(storePath string, manifest blockManifest) (*Block, error) {
+func loadBlock(storePath string, manifest BlockManifest) (*Block, error) {
 	b := &Block{
 		ID:        manifest.ID,
 		Name:      manifest.Name,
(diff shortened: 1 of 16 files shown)


scottjab:
Most of the issues I had with individual commits were addressed as a whole. 

I think the changes here are good.

LGTM

colinmarc:
This was baking in QA and I caught a deadlock - fixed in ded368d. 

PTAL, @scottjab?

colinmarc:
Ok, pushed one more fix. It's pretty subtle, but it was causing sequins to load one extra version than it needed to in some edge cases.

scottjab:
Let's let this bake a bit on QA before we go to prod, but besides that.

:shipit:

PR #63, opened 2016-10-31
Make port consistent throughout "Getting Started"

jdelaney-stripe:
An example uses port `9590` when Sequins was started up on port `9599` in an earlier example.

r? @colin-stripe


colin-stripe:
ha! amazing ❤️

PR #65, opened 2016-11-03
Cache old partition states incase of ZooKeeper failure.

scottjab:
Sequins will start 501ing if zookeeper has a blip.  This PR attempts to cache partition state between blips. 

r? @colinmarc 
cc? @charleseff

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -25,6 +25,7 @@ type partitions struct {
 	selected        map[int]bool
 	local           map[int]bool
 	remote          map[int][]string
+	old             []map[int][]string
 	numMissing      int
 	ready           chan bool
 	readyClosed     bool
@@ -44,6 +45,7 @@ func watchPartitions(zkWatcher *zkWatcher, peers *peers, db, version string, num
 		replication:   replication,
 		local:         make(map[int]bool),
 		remote:        make(map[int][]string),
+		old:           make([]map[int][]string, 1024),
 		ready:         make(chan bool),
 	}
 
@@ -128,6 +130,12 @@ func (p *partitions) updateRemotePartitions(nodes []string) {
 			remote[partition] = append(remote[partition], host)
 		}
 	}
+
+	// Keep track of old peers in case zookeeper goes away.
+	if len(p.old >= 1024) {
+		p.old = p.old[:1024]
+	}
+	p.old = append(remote, p.old)
 
 	p.remote = remote
 	p.updateMissing()
@@ -231,6 +239,11 @@ func (p *partitions) getPeers(partition int) []string {
 
 	peers := make([]string, len(p.remote[partition]))
 	copy(peers, p.remote[partition])
+
+	// Append old peers to peer list, in case of Zookeeper issues.
+	for _, oldPeer := range p.old {
+		peers = append(peers, oldPeer[partition])
+	}
 	return peers
 }
 


partitions.go
```diff
@@ -129,6 +131,26 @@ func (p *partitions) updateRemotePartitions(nodes []string) {
...
+		copy(newPartition, partition)
+
+		unDedupedPartition := append(newPartition, p.old[partitionId]...)
+		found := map[string]bool{}
+
+		// Shitty dedupe, iterate though the remote peers
+		for _, node := range unDedupedPartition {
+			if !found[node] {
+				found[node] = true
+				p.old[partitionId] = append([]string{node}, p.old[partitionId]...)
```
colinmarc: It'd probably be slightly more efficient to append rather than prepend, and then truncate as part of the dedupe (working backwards). Even better, though, would be to just have a separate `dedupe` method.

colinmarc: Also, this always adds the nodes to this list, when we just want to add them if they got removed above.

partitions.go
```diff
@@ -231,6 +253,12 @@ func (p *partitions) getPeers(partition int) []string {
 
 	peers := make([]string, len(p.remote[partition]))
 	copy(peers, p.remote[partition])
+
+	oldPeers := make([]string, 1024)
+	copy(oldPeers, p.old[partition])
+	// Append old peers to peer list, in case of Zookeeper issues.
```
colinmarc: Again, I don't think this comment tells the reader much (if they don't already know what's up).

partitions.go
```diff
@@ -231,6 +253,12 @@ func (p *partitions) getPeers(partition int) []string {
...
+	peers = append(peers, oldPeers...)
```
colinmarc: This means the dead peers will get shuffled into the live peers by the caller of this code. I think it'd be better if FindPeers return shuffled live peers + dead peers at the end.

partitions.go
```diff
@@ -129,6 +131,26 @@ func (p *partitions) updateRemotePartitions(nodes []string) {
...
 
+	// Keep track of old peers in case zookeeper goes away.
+	for partitionId, partition := range remote {
+		newPartition := make([]string, len(partition))
+		copy(newPartition, partition)
+
+		unDedupedPartition := append(newPartition, p.old[partitionId]...)
+		found := map[string]bool{}
+
+		// Shitty dedupe, iterate though the remote peers
```
colinmarc: Language!

partitions.go
```diff
@@ -111,6 +113,19 @@ func (p *partitions) updateLocalPartitions(local map[int]bool) {
 	}
 }
 
+func (p *partitions) deDupe(nodes []string) []string {
```
colinmarc: just call this `dedupe`

partitions.go
```diff
@@ -111,6 +113,19 @@ func (p *partitions) updateLocalPartitions(local map[int]bool) {
...
+	found := map[string]bool{}
+	dedupedNodes := make([]string, len(nodes))
```
colinmarc: this should be `make([]string, 0, len(nodes))` or you're going to have extra empty values at the end

partitions.go
colinmarc: also there's no reason for it to be on `p`: it can just be `func dedupe(s []string) []string`

partitions.go
```diff
@@ -129,6 +144,17 @@ func (p *partitions) updateRemotePartitions(nodes []string) {
 		}
 	}
 
+	for partitionId, partition := range remote {
```
colinmarc: this still isn't checking that the remote peer actually disappeared (ie, that it used to be in p.remote and now is not)

partitions.go
```diff
@@ -129,6 +144,17 @@ func (p *partitions) updateRemotePartitions(nodes []string) {
...
+		newPartition := make([]string, len(partition))
```
colinmarc: I'd call this `peers` I think

partitions.go
```diff
@@ -129,6 +144,17 @@ func (p *partitions) updateRemotePartitions(nodes []string) {
...
+		copy(newPartition, partition)
+
+		unDedupedPartition := append(newPartition, p.disappeared[partitionId]...)
```
colinmarc: you don't need the temp variable here.

partitions.go
```diff
@@ -221,17 +247,21 @@ func (p *partitions) partitionZKNode(partition int) string {
 }
 
 // getPeers returns the list of peers who have the given partition available.
```
colinmarc: You need to update the docstring. However, I still think it'd be cleaner to just return one list, and consider the peers "ordered" by priority. That gives us the freedom to actually add in priority later, if we want to do some light downweighting of peers from which we've seen errors.

colinmarc:
Speaking of the build failing, this would be great with some tests!

serve.go
```diff
@@ -120,12 +120,12 @@ func (vs *version) serveError(w http.ResponseWriter, key string, err error) {
 	w.WriteHeader(http.StatusInternalServerError)
 }
 
-func shuffle(vs []string) []string {
+func shuffle(vs []string, disappeared []string) []string {
```
colinmarc: This will return an arbitrary number of deleted nodes in addition to the live ones, which will change the behavior of proxy subtly. I think we want to try just <replication factor> peers, where dead peers are only considered if there aren't enough live peers to fulfill the replication factor.

PR #67, opened 2016-11-07
Fix a bug where 404s are turned into 500s

colinmarc:
In 38274f, we made the BlockStore object a lot more dumb; in
particular, we removed any knowledge it has about which partitions are
local (that state is instead stored on the partitions object now).

That means that if we don't have the partition in the block store,
it's probably because we never got any keys for that partition, not
anything broken. This might happen if, for instance, there was a file
with no keys in the input data.

r? @scottjab

### DIFF
--- a/blocks/block_store.go
+++ b/blocks/block_store.go
@@ -8,7 +8,6 @@ import (
 )
 
 var ErrNoManifest = errors.New("no manifest file found")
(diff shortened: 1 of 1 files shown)


colinmarc:
(I also did some drive-by refactoring of the method, obviously)

scottjab:
LGTM

colinmarc:
Travis is behind, but I ran the test locally, so I'm gonna merge.

PR #70, opened 2017-03-21
Bump golang version to 1.7.5

scottjab-stripe:
We have been seeing many dns failures that look a lot like this bug:
https://github.com/golang/go/issues/16865

Sequins was running on the release version of go 1.7, this PR bumps it to the most recent stable version of go 1.7.5

r? @evan-stripe


evan-stripe:
lgtm

PR #71, opened 2017-03-30
Store IP of sequins peers addresses in ZK.

scottjab-stripe:
When sequins proxies requests to other nodes in the cluster, it round trips to dns.  We have seen some issues with dns lookup failures when the cluster is really busy.  This will just store the peer IP addresses in ZK, which should reduce the DNS load. 

The downside is sequins _can_ lose peers, if ip addresses change it would require sequins to be restarted. 

r? @evan-stripe 
cc @bartle-stripe  @jdelaney-stripe @cff-stripe @sankari-stripe @as3richa-stripe @jbancroft-stripe @dusty-stripe @jbancroft-stripe

### DIFF
--- a/sequins.go
+++ b/sequins.go
@@ -133,14 +133,12 @@ func (s *sequins) initCluster() error {
 		return err
 	}
 
-	ip := ""
+	ip := hostname
 	ipAddresses, err := net.LookupHost(hostname)
 	if err != nil {
 		return err
 	}
-	if len(ipAddresses) != 1 {
-		ip = hostname
-	} else {
+	if len(ipAddresses) == 1 {
 		ip = ipAddresses[0]
 	}
 
--- a/sequins.go
+++ b/sequins.go
@@ -132,24 +132,24 @@ func (s *sequins) initCluster() error {
 	if err != nil {
 		return err
 	}
-	ipAddresses, err := net.LookupHost(hostname)
+	
 	ip := ""
+	ipAddresses, err := net.LookupHost(hostname)
 	if err	!= nil {
 		return err
 	}
-	if len(ipAddresses) < 1 {
+	if len(ipAddresses) != 1 {
 		ip = hostname
 	} else {
 		ip = ipAddresses[0]
 	}
+
 	routableIpAddress := net.JoinHostPort(ip, port)
 	routableAddress := net.JoinHostPort(hostname, port)
 	shardID := s.config.Sharding.ShardID
 	if shardID == "" {
 		shardID = routableAddress
 	}
-
-
 
 	peers := sharding.WatchPeers(zkWatcher, shardID, routableIpAddress)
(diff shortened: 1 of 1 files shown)


sequins.go
```diff
@@ -132,14 +132,26 @@ func (s *sequins) initCluster() error {
...
-
+	ipAddresses, err := net.LookupHost(hostname)
+	ip := ""
+	if err	!= nil {
+		return err
+	}
+	if len(ipAddresses) < 1 {
+		ip = hostname
+	} else {
+		ip = ipAddresses[0]
```
evan-stripe: Hmm. I suspect that in the event there are multiple IPs we want to just use the hostname?

sequins.go
```diff
@@ -132,14 +132,26 @@ func (s *sequins) initCluster() error {
...
+	routableIpAddress := net.JoinHostPort(ip, port)
 	routableAddress := net.JoinHostPort(hostname, port)
 	shardID := s.config.Sharding.ShardID
 	if shardID == "" {
 		shardID = routableAddress
 	}
 
-	peers := sharding.WatchPeers(zkWatcher, shardID, routableAddress)
+
+
```
evan-stripe: Seems like a lot of extra newlines :)

sequins.go
scottjab-stripe: Oh good idea,  Will only store IP if one is returned.

sequins.go
```diff
@@ -132,14 +132,26 @@ func (s *sequins) initCluster() error {
 	if err != nil {
 		return err
 	}
+	
+	ip := ""
+	ipAddresses, err := net.LookupHost(hostname)
+	if err	!= nil {
```
evan-stripe: Err, is that a tab or something? Something looks off with the formatting

(Do we validate that sequins code is gofmt'd?)

scottjab-stripe: My editor should go fmt, but I guess it didn't pushing change.  We don't currently validate gofmt.

sequins.go
```diff
@@ -133,13 +133,25 @@ func (s *sequins) initCluster() error {
...
+	ip := ""
+	ipAddresses, err := net.LookupHost(hostname)
+	if err != nil {
+		return err
+	}
+	if len(ipAddresses) != 1 {
+		ip = hostname
+	} else {
+		ip = ipAddresses[0]
+	}
```
evan-stripe: Oh, and I'm not sure why I missed this earlier (sorry), but I think this can be slightly better phrased as 

```go
ip := hostname
ipAddress, err := net.LookupHost(hostname)
if err != nil {
  return err
}
if len(ipAddresses) == 1 {
  ip = ipAddress[0]
}
```

evan-stripe:
One last nit, but either with that change or without
LGTM

PR #75, opened 2017-06-28
Filter out DBs and versions with invalid characters

evan-stripe:
We've seen problems from versions that inadvertently included special
characters (like '{'). To avoid problems, skip over anything that's
not alphanumeric.

r? @scottjab 
cc @stripe/storage


scottjab-stripe:
LGTM

PR #76, opened 2017-07-21
Close the body of any outstanding requests.

bartle-stripe:
Even if you cancel a request, it's still the callers responsibility to close the response Body if `http.Client.Do` returns a non-nil error. Cancelling a request while it is in the process of streaming a response merely prevents new data from being added to the response Body, but the Body itself remains open.

cc @stripe/storage

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -52,7 +52,22 @@ func (vs *version) proxy(r *http.Request, peers []string) (*http.Response, strin
 	// means it's canceled almost immediately.
 	// defer cancel()
 
+	// Per the documentation for `http.Client.Do`, if that returns successfully,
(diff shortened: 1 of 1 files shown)


proxy.go
```diff
@@ -53,6 +53,18 @@ func (vs *version) proxy(r *http.Request, peers []string) (*http.Response, strin
 	// defer cancel()
 
 	outstanding := 0
+	defer func() {
```
scottjab-stripe: Small nit, I would have a comment explaining why we are doing this.

bartle-stripe: Done.

scottjab-stripe:
LGTM

PR #78, opened 2017-07-24
Update aws-sdk to version 1.10.14

jerry-stripe:
An old aws-sdk is causing some issues with s3 testing, particularly when calling creating new blobs with empty bodies which should work, but currently doesn't. (related: https://github.com/aws/aws-sdk-go/issues/138)

r? @scottjab-stripe


scottjab-stripe:
LGTM

PR #79, opened 2017-07-24
Add healthcheck endpoints.

jerry-stripe:
Add a `/healthcheck` and `/healthz` endpoints that checks all of the versions of the databases of the given host and returns a `200` if they are all healthy and a `404` otherwise.

r? @scottjab-stripe 
cc? @wei-stripe @jbancroft-stripe @dusty-stripe @bartle-stripe @ikdc-stripe 

WIP


jerry-stripe:
I'm switching the response to be a JSON body so this is actually going to change in a bit

status.go
```diff
@@ -67,6 +67,44 @@ const (
...
+					ok = false
+				}
+			}
+		}
+	}
+
+	if ok {
+		w.WriteHeader(http.StatusOK)
+	} else {
+		w.WriteHeader(http.StatusServiceUnavailable)
```
scottjab-stripe: I think a 404 or a 410 (gone) is probably more appropriate for failure.

status.go
```diff
@@ -67,6 +67,44 @@ const (
...
+	}
+
+	for _, status := range statuses {
+		fmt.Fprintln(w, status)
```
scottjab-stripe: Let's return json.

status.go
```diff
@@ -67,6 +67,44 @@ const (
...
+	for name, db := range s.dbs {
+		status.DBs[name] = copyDBStatus(db.status())
+	}
+
+	s.dbsLock.RUnlock()
+
+	ok := true
+	var statuses []string
+
+	// Iterate through all of the nodes of each database and ensure that
```
scottjab-stripe: Nodes as in computers in the database or block locally?  The health check should be about local health, not the state of the cluster as a whole.

sequins.go
```diff
@@ -303,6 +303,11 @@ func (s *sequins) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if r.URL.Path == "/healthcheck" || r.URL.Path == "/healthz" {
```
scottjab-stripe: We should update the docs that you cannot have a db named healthcheck or healthz.

status.go
```diff
@@ -67,6 +67,59 @@ const (
...
+			for nodeKey, nodeValue := range versionValue.Nodes {
+				_, exists := statuses[dbKey]
+				if !exists {
+					statuses[dbKey] = make(map[string]map[string]versionState)
+				}
+
+				_, exists = statuses[dbKey][versionKey]
+				if !exists {
+					statuses[dbKey][versionKey] = make(map[string]versionState)
+				}
```
jerry-stripe: I'm not super thrilled about this map code, but I wasn't sure of a much better way to do it 😕

status.go
```diff
@@ -67,6 +67,59 @@ const (
...
+	ok := true
+	statuses := make(map[string]map[string]map[string]versionState)
+
+	// Iterate through all of the nodes of the local databases and ensure
+	// that the versions they own are all marked as available. A single
+	// version being in an unvailable state results in an error response.
+	for dbKey, dbValue := range status.DBs {
+		for versionKey, versionValue := range dbValue.Versions {
+			for nodeKey, nodeValue := range versionValue.Nodes {
+				_, exists := statuses[dbKey]
```
scottjab-stripe: small nit s/exists/ok  
It is go convention to call the second return from a map "ok"

scottjab-stripe:
Except for the small nit, I wonder if it is possible to write a test for this, but I don't know if that should be a blocker.

jerry-stripe:
I agree with testing, at the very least to make sure that the appropriate status code is returned. I wrote tests for that at least, I could add checks to make sure that the JSON response is valid as well?

jerry-stripe:
There is definitely a race condition somewhere, as the tests pass 3/4 of the time without changing anything. I suspect that maybe a test sequins instance interferes, so I specified new ports.

scottjab-stripe:
:shipit:

PR #81, opened 2017-07-26
add retries to s3backend::Open function

jerry-stripe:
Add retry functionality to the `Open` function for the s3 backend so that when calls to `GetObject` fail due to latency in updating the keys, it will be automatically handled.

r? @scottjab-stripe 
cc? @wei-stripe @jbancroft-stripe @dusty-stripe @bartle-stripe @ikdc-stripe

### DIFF
--- a/backend/s3_backend.go
+++ b/backend/s3_backend.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3iface"
 )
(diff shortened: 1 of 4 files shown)


backend/s3_backend.go
```diff
@@ -6,22 +6,26 @@ import (
...
 )
 
 type S3Backend struct {
-	bucket string
-	path   string
-	svc    *s3.S3
+	bucket     string
+	path       string
+	maxRetries int
+	svc        s3iface.S3API
```
jerry-stripe: I generalized to the interface here so that I could make a mock struct for testing.

backend/s3_backend.go
```diff
@@ -161,6 +165,15 @@ func (s *S3Backend) Open(db, version, file string) (io.ReadCloser, error) {
 	}
 	resp, err := s.svc.GetObject(params)
 
+	// If the download failed, retry maxRetries number of times with an
```
scottjab-stripe: I feel like we should check to see what the error is, before we start retrying. If we are getting a 400 level response code from s3 we should not retry.  Also adding a metric here for s3 retries might be a good idea.

config.go
```diff
@@ -95,6 +96,7 @@ func defaultConfig() sequinsConfig {
 		S3: s3Config{
 			Region:          "",
 			AccessKeyId:     "",
+			MaxRetries:      2,
```
scottjab-stripe: Small nit, I thing retrying 3 times is better, mostly for consistency with other retries in sequins.

backend/s3_backend.go
jerry-stripe: Oh yeah that makes sense - wouldn't we also want to retry on NoSuchKey? I don't exactly remember the particular problem that motivated this change but I thought that it had to do with the fact that S3 had sometimes not updated the set of keys available after writes occurred.

scottjab-stripe:
:shipit:

PR #83, opened 2017-08-02
Add GRPC skeleton interface

djudd-stripe:
cc @scottjab


djudd-stripe:
r? @scottjab-stripe 

I'm not super high confidence in this implementation, but I think it's ready for review. Two questions:
- Does the path to install protoc for builds look reasonable, or is it too high-friction for people on an unsupported (by us) OS?
- Is using the "standard" GRPC healthcheck protos actually a good idea? Is there any real-world tooling that cares about this? I was disappointed to see that in the Ruby client, you end up accessing the resulting endpoint via a separate client object, not under the Sequins namespace.

PR #84, opened 2017-08-02
reduce number of testing files, delete old files in S3 tests

jerry-stripe:
**Summary**

This reduces the number of tests sequence files and makes sure to delete previous files in the S3 bucket before performing S3 tests.

**Motivation**

In OSX, the number of files descriptors available by default is fairly low and this is starting to create issues for the Mac tests in Travis.

This also revealed the fact that the tests never clear out the s3 bucket used for testing, causing changes such as reducing the number of testing files to cause some tests to fail.

**Test plan**

Unit tests

Undo the commit

r? @scottjab-stripe 
cc @stripe/storage


scottjab-stripe:
LGTM

PR #85, opened 2017-08-02
add retries to s3backend::Open function

jerry-stripe:
Add retry functionality to the `Open` function for the s3 backend so that when calls to `GetObject` fail due to latency in updating the keys, it will be automatically handled.

r? @scottjab-stripe 
cc? @stripe/storage


scottjab-stripe:
LGTM

PR #86, opened 2017-08-03
Use `log.Printf` instead of `log.Println`.

bartle-stripe:
r? @scottjab-stripe 
cc @stripe/storage


evan-stripe:
Does govet catch this?

bartle-stripe:
It does! We should really run it as part of CI.

bartle-stripe:
Superseded by https://github.com/stripe/sequins/pull/87

PR #88, opened 2017-08-04
Hold `w.lock` when running zk hooks.

bartle-stripe:
This addresses the following race:
* thread-1 (in `run`): call `reconnect`
* thread-2: call `CreateEphemeral`
* thread-1: call `runHooks`

`CreateEphemeral` will create the node (which will succeed because thread-1 just finished connecting to zookeeper), and then add it to `ephemeralNodes`.  `runHooks` will then try to re-add the node, which will fail, triggering a zookeeper reconnection.

r? @scottjab 
cc @stripe/storage

### DIFF
--- a/zk/watcher.go
+++ b/zk/watcher.go
@@ -121,9 +121,16 @@ func (w *Watcher) reconnect() error {
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


scottjab-stripe:
LGTM

PR #89, opened 2017-08-04
Ignore ZCLOSING errors when creating ephemeral nodes and watches.

bartle-stripe:
It's safe to ignore these, since `runHooks` will recreate the ephemeral
nodes and watches as soon as we've reconnected. By not ignoring these,
we were triggering "Disconnecting from zookeeper because of error" in
the `run` loop.

r? @scottjab 
cc @stripe/storage


scottjab-stripe:
LGTM

PR #90, opened 2017-08-04
If we acquire `w.lock` and `w.hooksLock`, acquire the former first.

bartle-stripe:
This avoids this deadlock:
thread-1: acquire w.lock
thread-2: acquire w.hooksLock
thread-1: acquire w.hooksLock (blocks on thread-2)
thread-2: acquire w.lock (blocks on thread-1)
<deadlock>

r? @scottjab 
cc @stripe/storage


scottjab-stripe:
LGTM

PR #92, opened 2017-08-07
Begin to integrate Datadog to report metrics.

jerry-stripe:
**Summary**

This adds reporting to Datadog for two metrics:
1. Amount of time items take to download from S3
2. Amount of items waiting to be downloaded from S3 at a given time

**Motivation**

We currently don't have much visibility into metrics such as S3 download times or how backed up the download queue may be at any given time. This helps solve that.

**Test plan**

Automated tests

Revert commit

r? @scottjab-stripe 
cc? @stripe/storage

### DIFF
--- a/build.go
+++ b/build.go
@@ -80,7 +80,12 @@ func (vs *version) addFiles(partitions map[int]bool) error {
 	}
 
 	// TODO: parallelize files?
-	for _, file := range vs.files {
+	for i, file := range vs.files {
+		if vs.stats != nil {
+			remaining := float64(len(vs.files) - i - 1)
+			vs.stats.Gauge("s3.queue_depth", remaining, []string{vs.db.name}, 1)
+		}
+
 		select {
 		case <-vs.cancel:
 			return errCanceled
@@ -97,6 +102,14 @@ func (vs *version) addFiles(partitions map[int]bool) error {
 }
 
 func (vs *version) addFile(file string, partitions map[int]bool) error {
+	if vs.stats != nil {
+		start := time.Now()
+		defer func() {
+			duration := time.Since(start)
+			vs.stats.Timing("s3.download_duration", duration, []string{vs.db.name}, 1)
+		}()
+	}
+
 	disp := vs.sequins.backend.DisplayPath(vs.db.name, vs.name, file)
 	log.Println("Reading records from", disp)
 
--- a/build.go
+++ b/build.go
@@ -83,7 +83,8 @@ func (vs *version) addFiles(partitions map[int]bool) error {
 	for i, file := range vs.files {
 		if vs.stats != nil {
 			remaining := float64(len(vs.files) - i - 1)
-			vs.stats.Gauge("s3.queue_depth", remaining, []string{vs.db.name}, 1)
+			tags := []string{fmt.Sprintf("db:%s", vs.db.name)}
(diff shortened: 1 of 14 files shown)


build.go
```diff
@@ -97,6 +102,16 @@ func (vs *version) addFiles(partitions map[int]bool) error {
 }
 
 func (vs *version) addFile(file string, partitions map[int]bool) error {
+	if vs.stats != nil {
+		start := time.Now()
+
+		defer func() {
+			duration := time.Since(start)
+			vs.stats.Timing("s3.download_duration", duration, nil, 1)
```
scottjab-stripe: Add a tag for which db.

build.go
```diff
@@ -79,6 +79,11 @@ func (vs *version) addFiles(partitions map[int]bool) error {
 		return nil
 	}
 
+	if vs.stats != nil {
+		remaining := len(vs.files)
+		vs.stats.Count("s3.queue_depth", int64(remaining), nil, 1)
```
scottjab-stripe: tags

jerry-stripe:
New to dealing with observability-related things so I would love it if someone from @stripe/observability would take a look as well!

r? @stripe/observability

build.go
```diff
@@ -79,6 +79,11 @@ func (vs *version) addFiles(partitions map[int]bool) error {
 		return nil
 	}
 
+	if vs.stats != nil {
+		remaining := len(vs.files)
+		vs.stats.Count("s3.queue_depth", int64(remaining), []string{vs.db.name}, 1)
```
krisreeves-stripe: For queue depth you probably want a gauge

build.go
```diff
@@ -97,6 +102,16 @@ func (vs *version) addFiles(partitions map[int]bool) error {
 }
 
 func (vs *version) addFile(file string, partitions map[int]bool) error {
+	if vs.stats != nil {
+		start := time.Now()
+
+		defer func() {
+			duration := time.Since(start)
+			vs.stats.Timing("s3.download_duration", duration, []string{vs.db.name}, 1)
+			vs.stats.Decr("s3.queue_depth", []string{vs.db.name}, 1)
```
krisreeves-stripe: Similarly, here, don't emit a decrement at all; just emit the full size of the value as a gauge when it changes/periodically/whatever makes sense. Making increments and decrements line up over time in DD dashboards is near impossible, and they don't come across as expected usually

config.go
```diff
@@ -128,6 +133,9 @@ func defaultConfig() sequinsConfig {
 				SecretAccessKey: "",
 			},
 		},
+		Datadog: datadogConfig{
+			Bind: "localhost:8200",
```
lopopolo-stripe: This seems like an unconventional usage of `bind`. Are you creating a socket on this address? Or are you sending packages/datagrams to it?

build.go
jerry-stripe: Oh yes, you're totally right. I thought I would have to take this approach since the encapsulating function is called as a goroutine and I thought that using gauge would introduce the possibility threads writing over each other. As it turns out a lock is taken out and I wasn't tagging metrics so this is actually not an issue.

config.go
jerry-stripe: Sending datagrams -- I guess that something like `url` is more appropriate?

scottjab-stripe:
:shipit:

PR #95, opened 2017-08-15
tweak sequins tags

jerry-stripe:
Change the tags used for sequins reporting to datadog to be more consistent with what is currently in use.

r? @scottjab-stripe


scottjab-stripe:
Shouldn't this just be pushed to your current branch?

jerry-stripe:
You know what, I just assumed it wasn't kosher to add to merged branches, but I should probably just do that.

scottjab-stripe:
👍

PR #96, opened 2017-08-15
s/cluster/sequins for tagging

jerry-stripe:
Change `cluster` tag name to `sequins` in accordance with current standards.

r? @scottjab-stripe


scottjab-stripe:
LGTM

PR #97, opened 2017-08-16
Establish a minimum replication level before switching to a new version

vasi-stripe:
This prevents errors in the following case:

* A new version becomes available
* Nodes start reading the files from the new version, but some nodes are faster than others
* At some point, at least one replica of each partition exists, and the new version goes active
    * Because of variable speed, many partitions still have only one replica
* A node goes down, and parts of the new version are unavailable. Clients get errors
* Eventually other nodes read enough data, and things settle down

I think my changes are roughly correct, however this is really hard to test. I have a partial test case, but I had to add some grotty code to simulate crashes, and I'm still getting sporadic failures.

r? @bartle-stripe

### DIFF
--- a/cluster_test.go
+++ b/cluster_test.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"syscall"
 	"testing"
 	"time"
@@ -91,7 +92,21 @@ func newTestCluster(t *testing.T) *testCluster {
 	}
 }
 
-func (tc *testCluster) addSequins() *testSequins {
+type configOption func(*sequinsConfig)
+
+func repl(r int) configOption {
+	return func(config *sequinsConfig) {
+		config.Sharding.Replication = r
+	}
+}
+
+func minRepl(r int) configOption {
+	return func(config *sequinsConfig) {
+		config.Sharding.MinReplication = r
+	}
+}
+
+func (tc *testCluster) addSequins(opts ...configOption) *testSequins {
 	port := zktest.RandomPort()
 	path := filepath.Join(tc.source, fmt.Sprintf("node-%d", port))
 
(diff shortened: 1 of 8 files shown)


cluster_test.go
```diff
@@ -13,6 +13,8 @@ import (
 	"testing"
 	"time"
 
+	"strconv"
```
scottjab-stripe: nit, this should be up with the std lib packages, sometimes go imports is stupid.

vasi-stripe: I actually have a little tool for that! https://git.corp.stripe.com/vasi/gogroup

It finds a few things in sequins :(
```
> 
find . -name vendor -prune -o -name \*.go -exec gogroup -order std,other,prefix=github.com/stripe/sequins/ '{}' +

./cluster_test.go:20: Import in incorrect group at "github.com/stripe/sequins/zk/zktest"
./hdfs_sequins_test.go:13: Import in incorrect group at "github.com/stripe/sequins/backend"
./main.go:15: Import in incorrect group at "github.com/stripe/sequins/backend"
./s3_sequins_test.go:19: Import in incorrect group at "github.com/stripe/sequins/backend"
./zk/watcher_test.go:11: Import in incorrect group at "github.com/stripe/sequins/zk/zktest"
```

scottjab-stripe: Super cool!

vasi-stripe:
Ok, I think this is ready for review. @bartle-stripe PTAL

sharding/partitions.go
```diff
@@ -233,19 +237,22 @@ func (p *Partitions) updateRemote(nodes []string) {
 }
 
 func (p *Partitions) updateMissing() {
-	// Check for each partition. If every one is available on at least one node,
+	// Check for each partition. If every one is available on at least minReplication node,
 	// then we're ready to rumble.
 	missing := 0
```
bartle-stripe: Maybe rename this to `underReplicated` (there's a few occurrences/comments referencing `missing`/`Missing` in this file).

sharding/partitions.go
```diff
@@ -233,19 +237,22 @@ func (p *Partitions) updateRemote(nodes []string) {
...
+			replication += len(remotes)
 		}
 
-		missing += 1
+		if replication < p.minReplication {
+			missing += 1
+		}
 	}
 
 	p.numMissing = missing
```
bartle-stripe: I think we could actually delete `p.numMissing`? The only thing that reads it is `Missing()`, which isn't used anywhere.

bartle-stripe:
LGTM

Couple nits, which you can feel free to ignore or follow up on later.

vasi-stripe:
Renamed `missing`, and rebased.

PR #98, opened 2017-08-16
Ignore intellij projects and sequins configs.

scottjab-stripe:
I accidentally wiped my sequins config and project file with a `git clean` let's just ignore that stuff. 

r? @jerry-stripe 
cc @stripe/storage


jerry-stripe:
LGTM

PR #100, opened 2017-08-16
DB load metric.

scottjab-stripe:
Emit a metric for how long it took the db version to download.

r? @jerry-stripe 
cc @stripe/storage

### DIFF
--- a/status.go
+++ b/status.go
@@ -388,8 +388,8 @@ func (vs *version) setState(state versionState) {
 
 			if vs.stats != nil {
 				tags := []string{fmt.Sprintf("sequins_db:%s", vs.db.name)}
-				span := vs.available.Sub(vs.created)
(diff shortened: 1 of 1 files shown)


status.go
```diff
@@ -385,6 +385,12 @@ func (vs *version) setState(state versionState) {
 		vs.state = state
 		if state == versionAvailable {
 			vs.available = time.Now()
+
+			if vs.stats != nil {
+				tags := []string{fmt.Sprintf("sequins_db:%s", vs.db.name)}
+				span := vs.available.Sub(vs.created)
```
jerry-stripe: Actually turns out I called this `duration` instead of `span`, since I think spans imply tracing contexts, but I don't think it really matters.

jerry-stripe:
LGTM

status.go
scottjab-stripe: Will fix for consistency.

PR #103, opened 2017-08-25
Add logging of datasets

scottjab-stripe:
This adds a log line that reports how many files and how big a given database version has.

r? @jerry-stripe 
cc @stripe/storage


jerry-stripe:
lgtm

Do you think it's worth reporting this in datadog too?

scottjab-stripe:
Hmm maybe without the version? I think including the version as a tag would be too much tag cardinality.

PR #104, opened 2017-08-29
Double Sequins' timeout for ZK sessions and Ignore ErrNodeExists 

scottjab-stripe:
This PR addresses two issues:

1. Sequins is racing the ZK server's timeouts, Sequins will timeout the goroutine before ZK times itself out.  ZK then will still be timing out when Sequins reconnects.  This doubling the timeout on the goroutine will let ZK handle the timeout of its sockets and znodes.  We should still see the timeout happen though a disconnection if this happens.  The timeout on the goroutine is mostly for thread lock safety. 

2. When sequins creates parent znodes sometimes it will fail if the parents are already created, this leads to a reconnect.  We should ignore znode exists errors when we are creating the parents.

r?  @jerry-stripe 
cc @stripe/storage

### DIFF
--- a/zk/watcher.go
+++ b/zk/watcher.go
@@ -331,9 +331,7 @@ func (w *Watcher) watchChildren(node string, wn watchedNode) error {
 			case reconnecting = <-wn.cancel:
 				return
 			case ev := <-events:
-
 				if ev.Err != nil {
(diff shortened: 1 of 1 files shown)


jerry-stripe:
lgtm

PR #107, opened 2017-09-06
Use smallest missing shardID if none is given

jdelaney-stripe:
Tested with the following procedure:
1. Remove `shard_id` from the configuration file
2. Restart Sequins and make sure it does smallest missing to figure out the correct `shardID`
3. Restart Sequins again and make sure it reads from the `self_assigned_id` file
4. Change the `self_assigned_id` file to be "0\n" and then restore the configuration file to include `shard_id` again
5. Restart Sequins again and make sure it uses the configuration file over the `self_assigned_id` file

I tested this on tha node with the lowest shardID, the highest shardID, and a shardID in the middle.

This has a fair amount of logging lines. Are there any you think I should remove?

r? @scottjab-stripe

### DIFF
--- a/sequins.go
+++ b/sequins.go
@@ -18,7 +18,6 @@ import (
 
 	"github.com/DataDog/datadog-go/statsd"
 	"github.com/nightlyone/lockfile"
-	gozk "github.com/samuel/go-zookeeper/zk"
 	"github.com/tylerb/graceful"
 
 	"github.com/stripe/sequins/backend"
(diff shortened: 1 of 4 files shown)


sequins.go
```diff
@@ -159,11 +161,52 @@ func (s *sequins) initCluster() error {
...
 	if shardID == "" {
-		shardID = routableAddress
+		fp := filepath.Join(s.config.LocalStore, "self_assigned_id")
+		_, err := os.Stat(fp)
+		if err != nil {
+			if os.IsNotExist(err) {
+				log.Print("No self-assigned ID file found")
+
+				lock := zkWatcher.CreateIDAssignmentLock()
+				lock.Lock()
```
bartle-stripe: Could we structure this in a way that we could add `defer lock.Unlock()`? Maybe move lines 165-209 to a separate function?

sequins.go
```diff
@@ -118,6 +119,53 @@ func (s *sequins) init() error {
 	return nil
 }
 
+func (s *sequins) joinCluster(zkWatcher *zk.Watcher, routableIpAddress string) (*sharding.Peers, error) {
+	lock := zkWatcher.CreateIDAssignmentLock()
```
jdelaney-stripe: I changed the locking to take out a much larger lock. Previously it just locked when a shardID wasn't provided and only locked until one was found. Now it locks until the cluster has been joined. This avoids the race condition of 2 Sequins nodes starting up, the first one choosing ID `X`, releasing the lock, and then the second one also choosing `X` while the first one is stuck on waiting for the cluster to stabilize [0]

[0] I realize this is very unlikely since the second one would also need to wait for the cluster to stabilize. But if, for some reason, we ever changed this logic or the two `WaitToConverge` times were different then this could be an issue

jdelaney-stripe:
It's worth noting the tests I conducted for this PR were done on 7d14c96. I will rerun those tests after this PR is approved and before it is merged in.

zk/watcher.go
```diff
@@ -210,6 +210,18 @@ Reconnect:
 	w.cancelWatches()
 }
 
+func (w *Watcher) CreateIDAssignmentLock() *zk.Lock {
+	node := path.Join(w.prefix, "id_assignment_lock")
+
+	w.lock.RLock()
+	defer w.lock.RUnlock()
+
+	w.hooksLock.Lock()
```
bartle-stripe: I don't think you need the `hooksLock`; that only protects `ephemeralNodes` and `watchedNodes`.

sharding/peers.go
```diff
@@ -33,6 +35,54 @@ type peer struct {
 	address string
 }
 
+func (p *Peers) SmallestMin() (string, error) {
```
bartle-stripe: Maybe `SmallestAvailableShardID`?

sharding/peers_test.go
```diff
@@ -0,0 +1,70 @@
+package sharding
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPeers_SmallestMin(t *testing.T) {
```
bartle-stripe: I don't think you generally use underscores in golang in this way.

sharding/peers.go
```diff
@@ -33,6 +35,54 @@ type peer struct {
...
+			return "", fmt.Errorf("Can't convert shardID %q to int", peer.shardID)
+		}
+
+		peerList[i] = num
+		i++
+	}
+
+	sort.Ints(peerList)
+
+	if peerList[0] != 1 {
```
bartle-stripe: I don't think you need this? If `peerList` starts with e.g. 2, you'll enter the loop and 2 != (0+1) and you'll return 0+1 as expected.

sharding/peers.go
```diff
@@ -33,6 +35,54 @@ type peer struct {
 	address string
 }
 
+func (p *Peers) SmallestMin() (string, error) {
+	if len(p.peers) == 0 {
```
bartle-stripe: Likewise, don't think you need this, sine for an empty list you'll return `prevNum+1`, and `prevNum` would be 0.

sharding/peers.go
```diff
@@ -33,6 +35,46 @@ type peer struct {
 	address string
 }
 
+func (p *Peers) SmallestAvailableShardID() (string, error) {
+	peerList := make([]int, len(p.peers), len(p.peers))
```
bartle-stripe: You can also just do `make([]int, 0, len(p.peers))` and `peerList = append(peerList, num)` so you don't need the `i` counter, but meh.

bartle-stripe:
lgtm

PR #108, opened 2017-09-08
Expose node's shardID in /healthcheck

jdelaney-stripe:
This is useful information to expose to users. It's also needed so that we can write automated tests for auto-assigning shardIDs.

r? @scottjab-stripe


scottjab-stripe:
LGTM

PR #109, opened 2017-09-08
Change port in guide from 9590 to 9599

jdelaney-stripe:
By default Sequins uses port 9599 and that's the port used in all examples in the guide so for user experience we should keep it consistent.

r? @jerry-stripe


jerry-stripe:
lgtm!

PR #110, opened 2017-09-13
Parallelize by individual file, rather than dataset, when downloading.

bartle-stripe:


bartle-stripe:
cc @stripe/storage

scottjab-stripe:
Really nice.

LGTM

PR #111, opened 2017-09-15
Refactor functional tests and have them run again

jdelaney-stripe:
### Summary

Functional tests were disabled in CI almost a year ago due to flakiness (62b72b01c27ccbd7e77202b240ac8b713bb2e4d9). This refactor reenables them and address the sources of flakiness:

* There was no checking to see if Sequins actually starts up when `start` is called. This makes it possible for Sequins to randomly choose a port that's in use, fail to start up, but have the tests continue (and then fail) [0]
* Sometimes Sequins would start up too quickly and the state polling would miss states like `down`. Because this PR changes it to better control the startup process we don't need to poll for the initial `down` state.
* The `testClient` timeout was pretty low and would sometimes cause issues.

### Testing Process

Testing this change has consisted of running the functional test suite 100 times locally and 60 times on Travis. None of these runs failed.

### Future Work

I think we can make the testing framework for functional tests even better. Right now they rely heavily on polling the current state of Sequins. This works pretty well, but is susceptible to timing issues (if Sequins is too fast the polling can miss a state). Instead we should do one of the following:
* Have Sequins output its state and then using the log as the source of progression.
* Assert the progression at certain times throughout the test where it's known to be stable.

### Footnotes

[0] This was actually a compounding issue. The test would fail with a `panic` so the `defer tc.tearDown()` functions would never be run, leaving a lot of Sequins nodes running in the background. This would increase the likelihood of reusing a port in future tests.

Thanks to @jeremy-stripe for help with figuring out what was causing issues in these tests.

r? @scottjab-stripe 
cc @vasi-stripe @bartle-stripe @jerry-stripe @dusty-stripe

### DIFF
--- a/cluster_test.go
+++ b/cluster_test.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strconv"
+	"sync"
 	"syscall"
 	"testing"
 	"time"
@@ -27,7 +28,6 @@ type testVersion string
 
 const dbName = "db"
 const (
-	start     testVersion = "START"
 	noVersion testVersion = "NONE"
 	v1        testVersion = "v1"
 	v2        testVersion = "v2"
@@ -56,6 +56,7 @@ type testSequins struct {
 	testClient          *http.Client
 	expectedProgression []testVersion
 
+	cmdError    chan error
 	process     *exec.Cmd
 	log         *os.File
 	progression chan testVersion
@@ -78,7 +79,7 @@ func newTestCluster(t *testing.T) *testCluster {
 	// We have a specific transport to the client, so it doesn't try to reuse
 	// connections between tests
 	var testClient = &http.Client{
-		Timeout:   500 * time.Millisecond,
+		Timeout:   5000 * time.Millisecond,
 		Transport: &http.Transport{},
 	}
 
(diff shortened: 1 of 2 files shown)


cluster_test.go
```diff
@@ -590,12 +596,11 @@ func TestClusterNodeWithoutData(t *testing.T) {
 
 	// Because it's behind, it's expected that the first node will flap when it
 	// proxies requests and gets v2 from peers.
-	// tc.sequinses[0].expectProgression(down, noVersion, v1, v3)
-	tc.sequinses[1].expectProgression(down, noVersion, v1, v2, v3)
-	tc.sequinses[2].expectProgression(down, noVersion, v1, v2, v3)
+	// tc.sequinses[0].expectProgression(v1, v3)
```
scottjab-stripe: is there a reason this is commented out?

jdelaney-stripe: My PR didn't comment this out, it was already like that (0eda089523d171507373fc68cf8d2b9eeb801473). I think the comment above it is saying that this is what we would test, but we can't because it will be flappy (expected).

jdelaney-stripe:
Adding `approved` label myself since @stripe-ci didn't

PR #112, opened 2017-09-15
Add functional tests for shardID auto-assignment

jdelaney-stripe:
Also changes Sequins shardID to start from 1 rather than 0 when using the config file to set the `shardID`. This better matches Sequins behavior and expectations.

r? @bartle-stripe 
cc @scottjab-stripe @bartle-stripe @dusty-stripe


bartle-stripe:
lgtm

PR #114, opened 2017-09-16
Output sequencefile header in logs.

bartle-stripe:
I think we'll want to export compression settings to datadog so that we can potentially alert/file tickets on uncompressed data, but this gives us something to start with.

r? @vasi-stripe 
cc @stripe/storage


vasi-stripe:
LGTM

PR #115, opened 2017-10-02
have healthcheck endpoint return 200 iff the latest versions are heal…

jerry-stripe:
**Summary**
This changes the healthcheck response such that 200 will only be returned if the _latest_ version of each db is available. Previously, it required that all versions of each db were available to return 200. This also includes more general information about the given node for each version as opposed to just the status.

**Motivation**
This makes it possible to simplify the `roll-sequins` golem as it will no longer have to duplicate this logic and instead just check for a 200 status code. Also prepares for status page changes by including partition/block information in the healthcheck.

**Test plan**
Automated tests

Revert commit

r? @scottjab-stripe
cc? @stripe/storage

### DIFF
--- a/status.go
+++ b/status.go
@@ -81,25 +81,35 @@ func (s *sequins) serveHealth(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("X-Sequins-Shard-ID", s.peers.ShardID)
 	}
 
-	allNodesAvailable := true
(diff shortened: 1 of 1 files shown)


scottjab-stripe:
So what happens when the cluster is loading new versions?

jerry-stripe:
Closing in favor of https://github.com/stripe/sequins/pull/118

PR #116, opened 2017-10-03
refactor test source directories

jerry-stripe:
**Summary**
Create the `test_databases` directory which contains the sources that unit tests will use. It currently contains the `healthy` and `unhealthy` sources representing a valid sequins source and a sequins source that will cause errors when loaded.

**Motivation**
Allows us to create more sources with different initial states without cluttering root sequins directory.

r? @scottjab-stripe 
cc? @stripe/storage


jerry-stripe:
Well I guess we have flaky tests or something because they pass now 😕 . PTAL @scottjab-stripe

scottjab-stripe:
LGTM

PR #117, opened 2017-10-05
redesign sequins dashboard, add version node state

jerry-stripe:
**Summary**
- Exposes information about the status of nodes within each version.
- Modifies the front page of the dashboard to show more general information while leaving the specific information for when you click through to a database. Only the amount of nodes in a given state will be revealed for each version on the front page.
- Implement more stripey design
- Add a new `ACTIVE` state on top of the `AVAILABLE` one. An `ACTIVE` version is the one that is actively being served, and is thus a stronger statement than `AVAILABLE` which only means that it is still capable of being served by sequins. For example, the latest version might be `AVAILABLE` but not `ACTIVE` for a given node until all of its peers also mark the version as `AVAILABLE`.

**Motivation**
It's hard to tell what the state of a given database is from the colors of the blocks alone. Adding more explicit labels and bolding certain aspects will hopefully make it easier to get information at a glance.

**Test plan**
This has already been deployed to dummysequins.

r? @scottjab-stripe 
cc? @stripe/storage

### DIFF
--- a/status.tmpl
+++ b/status.tmpl
@@ -3,16 +3,18 @@
     <style type="text/css">
       body {
         font-family: monospace;
+        color: #32325d;
         font-size: 12px;
         margin: 0px;
-        background-color: whitesmoke;
+        background-color: #f6f9fc;
       }
 
       div#header {
         width: 100%;
-        background-color: #2b83ba;
-        padding: 7px 20px;
-        color: white;
+        background-color: #525f7f;
+        padding: 12px 20px;
+        color: #f6f9fc;
+        box-shadow: 0 4px 6px rgba(50, 50, 93, 0.11), 0 1px 3px rgba(0, 0, 0, 0.08)
       }
 
       div#header > h1 {
@@ -24,11 +26,11 @@
       }
 
       div.db {
-        padding: 10px;
+        padding: 20px;
         margin: 0 auto 20px auto;
         display: table;
         background-color: white;
-        box-shadow: 5px 5px 10px 0px rgba(0,0,0,0.15);
+        box-shadow: 0 7px 14px rgba(50, 50, 93, 0.10), 0 3px 6px rgba(0, 0, 0, 0.08);
       }
 
       h2.dbname > a {
@@ -42,7 +44,7 @@
       }
 
       div.version {
-        border-top: 1px solid lightgray;
+        border-top: 1px solid #e6ebf1;
         padding: 10px 5px 10px 10px;
         display: block;
       }
@@ -53,12 +55,31 @@
 
       span.versionname {
         font-size: 20px;
+        font-weight: bold;
+      }
+
+      span.versionstate {
+        float: right;
+        font-size: 16px;
+        font-weight: bold;
+        padding-top: 0px;
+      }
+
+      span.versionstate > span.count {
+        margin-left: 10px;
+      }
+
+      span.versionstate > span.label {
+        padding: 0px 4px;
+        border-radius: 4px;
+        color: white;
+        margin-left: 4px;
       }
 
       span.versionpath {
-        font-size: 10px;
+        font-size: 12px;
         padding-left: 20px;
-        color: #888888;
+        color: #8898aa;
       }
 
       div.versionwrapper > div {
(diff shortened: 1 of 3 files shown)


jerry-stripe:
Also a look at what the front page would look like with a variety of node states:

screen shot 2017-10-05 at 12 08 41 pm

scottjab-stripe:
I am probably not the best person to review the webby bits, but the changes to sequins seem fine to me.

LGTM

jbancroft-stripe:
I know @dusty-stripe in a former life was a web developer 😃

jerry-stripe:
If that's the case, would love if he could take a look at it too :)

PTAL @dusty-stripe

status.tmpl
```diff
@@ -162,67 +222,108 @@
...
-                  (available since {{ $node.AvailableAt }})
-                {{ end }}
+        <div class="db">
+          <h2 class="dbname">/{{ $dbName }} {{ if gt (len $.DBs) 1 }}<a href="/{{ $dbName }}">&rarr;</a>{{ end }}</h2>
+          {{ range $versionName, $version := $db.Versions }}
+
+          <div class="version">
+            <span class="versionname">/{{ $versionName }}</span>
+
+            {{ if gt (len $.DBs) 1 }}
```
jerry-stripe: I should note that lines like these are a hacky result of the fact that sequins uses the same template to serve status pages for all databases and 1 database. I saw it used as a way to render the page a bit differently based on if we're looking at all databases v. just one so I did the same. Definitely not ideal though.

status.tmpl
```diff
@@ -162,67 +222,108 @@
           }
         }
       };
+
+      var countNodeStatuses = function() {
+        var dbs = {{ marshal . }};
+        var elements = document.getElementsByClassName('versionstate');
+        [].forEach.call(elements, function (elem) {
```
dusty-stripe: This is an interesting way to iterate an array. Is `elements` only array-like?

status.tmpl
dusty-stripe: Yup, that's pretty odd. As a simple clean-up, we could at least place a boolean on the view model stating if there's just 1 so you don't have to check the length of the list over and over.

status.tmpl
```diff
@@ -3,16 +3,18 @@
     <style type="text/css">
       body {
```
dusty-stripe: Not going to touch the presence of inline CSS and JavaScript. ;)

status.tmpl
```diff
@@ -162,67 +222,108 @@
...
+        var elements = document.getElementsByClassName('versionstate');
+        [].forEach.call(elements, function (elem) {
+          var dbName = elem.getAttribute('data-dbName');
+          var versionName = elem.getAttribute('data-versionName');
+
+          var statusCounts = {};
+          var nodes = dbs.dbs[dbName].versions[versionName].nodes;
+          Object.values(nodes).forEach(function (node) {
+            if (node.state in statusCounts) {
+              statusCounts[node.state]++
```
dusty-stripe: Missing semi-colon.

status.tmpl
```diff
@@ -162,67 +222,108 @@
...
+            } else {
+              statusCounts[node.state] = 1;
+            }
+          });
+
+          for (status in statusCounts) {
+            var count = document.createElement('span');
+            count.innerHTML = statusCounts[status].toString() + 'x';
+            count.classList.add('count')
```
dusty-stripe: semi-colon here too.

dusty-stripe:
Aside from a couple of missing semi-colons in the javascript and some concerns outside the scope of this PR, LGTM ;)

status.tmpl
jerry-stripe: Yeah it's an HtmlCollection, so this is the best alternative I could find to `Array.prototype.forEach.call...`

status.tmpl
jerry-stripe: So I tried looking a way around this and the view model is the sequins struct used in the actual gocode, so I could add a method to it, but it seemed strange to add a method that would be added just for the benefit of this single use case.

dusty-stripe: When utilized to it's fullest, the purpose of the view model is to remove as much logic from the view as possible, moving it to the controller, where it's more readily testable. I wouldn't add a method to the struct so much as a property calculated in the go code.

status.tmpl
```diff
@@ -162,67 +222,108 @@
...
-                  (available since {{ $node.AvailableAt }})
-                {{ end }}
+        <div class="db">
+          <h2 class="dbname">/{{ $dbName }} {{ if (isListView $.DBs) }}<a href="/{{ $dbName }}">&rarr;</a>{{ end }}</h2>
+          {{ range $versionName, $version := $db.Versions }}
+
+          <div class="version">
+            <span class="versionname">/{{ $versionName }}</span>
+
+            {{ if (isListView $.DBs) }}
```
jerry-stripe: @dusty-stripe I added the `isListView` function to the template since I just realized I could :P

It's not any more concise, but hopefully it conveys more meaning?

dusty-stripe: 👍

dusty-stripe:
LGTM

PR #118, opened 2017-10-05
include more info in healthcheck, return 200 if any versions available

jerry-stripe:
**Summary**
Include which partitions the node owns for each version in `/healthcheck` endpoint. Relax "healthiness" requirement by having `/healthcheck` return a `200` if any version is available in any database.

**Motivation**
Exposing partition ownership information in `/healthcheck` makes it easier to write the golem that will delete over-replicated partitions.

**Testing**
Unittests plus manual verification that the JSON response is as expected.

r? @scottjab-stripe 
cc? @stripe/storage


jerry-stripe:
fixed conflicts with new dashboard, PTAL @scottjab-stripe

scottjab-stripe:
LGTM

PR #119, opened 2017-10-10
actually use targetReplication field in versionStatus

jerry-stripe:
The `targetReplication` field was never being used, leading to the status page replication info being a bit less useful (i.e. xxx replication: xx/0). This sets it to `min_replication` from the config. Another option is setting it to just `replication` from the config if y'all think that's preferable.

r? @scottjab-stripe 
cc? @stripe/storage


scottjab-stripe:
LGTM

PR #120, opened 2017-10-12
Allow simultaneous adds to different partitions

vasi-stripe:
This lets downloads go several times faster on well-partitioned data. We can hit > 200 MB/s.

r? @bartle-stripe


bartle-stripe:
lgtm

PR #122, opened 2017-10-12
Use status 499 when client cancels request

jdelaney-stripe:
When the client cancels a request before it finishes we don't want to count that as an internal server error because it isn't. Instead we should use 499 to indicate that the client closed the request[0].

[0] Note that 499 is a nonstandard HTTP status code and not in the official spec.

r? @bartle-stripe 
cc @jerry-stripe @vasi-stripe @scottjab-stripe @dusty-stripe

### DIFF
--- a/serve.go
+++ b/serve.go
@@ -85,6 +85,11 @@ func (vs *version) serveProxied(w http.ResponseWriter, r *http.Request,
 		log.Printf("All peers timed out for /%s/%s (version %s)", vs.db.name, key, vs.name)
 		w.WriteHeader(http.StatusGatewayTimeout)
 		return
+	} else if err == errRequestCanceled {
+		// The connection was closed by the client. 499.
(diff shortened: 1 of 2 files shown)


serve.go
```diff
@@ -117,7 +117,11 @@ func (vs *version) serveNotFound(w http.ResponseWriter) {
 
 func (vs *version) serveError(w http.ResponseWriter, key string, err error) {
 	log.Printf("Error fetching value for /%s/%s: %s\n", vs.db.name, key, err)
-	w.WriteHeader(http.StatusInternalServerError)
+	if err == errRequestCanceled {
```
bartle-stripe: We already check some errors in `serveProxied` so I think it'd make more sense to add this there?

bartle-stripe: https://github.com/stripe/sequins/blob/master/serve.go#L77

bartle-stripe:
lgtm

PR #123, opened 2017-10-16
make partition replication information a bit more clear

jerry-stripe:
Update the partition replication information to be a bit more clear about how many partitions are in which replication states. Also make explicit what is the average v. target replication.

screen shot 2017-10-16 at 12 43 00 pm

r? @bartle-stripe 
cc? @stripe/storage


bartle-stripe:
lgtm

PR #124, opened 2017-10-19
replace hash ring partition assignment with sequential scheme

jerry-stripe:
**Summary**
Changes the partition assignment scheme to use a sequential assignment to nodes as opposed to a hash ring. A lot of lines where changed, but this is all mainly ripping out the shard assignment logic as it's no longer needed. The ordering for nodes is based on the hostname of that node now. While this means that a node can no longer "replace" a downed one by using the same shard ID, it turns out that in practice this never worked very well. Specific changes are as follows:
- Change partition assignment to use sequential ordering instead of hash ring.
- Remove Shard ID logic as it's no longer needed.
- Update documentation and tests to reflect removal of Shard ID.

**Motivation**
- Will distribute partitions amongst nodes much more uniformly.
- Will remove error-prone shard ID assignment logic which was prone to race conditions and complicated.

r? @scottjab-stripe 
cc? @stripe/storage


cluster_test.go
```diff
@@ -821,89 +837,9 @@ func TestClusterMin(t *testing.T) {
...
-		assert.True(t, found, "None of the nodes had shardID %q", shardID)
-	}
-}
-
-// Down each node in a cluster one at a time and then ensure shardID auto-assignment will use
-// the newly unused shardID.
-func TestClusterAutoAssignUnusedID(t *testing.T) {
+// TestClusterPartitionAssignment tests that the partitions are assigned to
+// nodes as expected since the assignments should be deterministic.
+func TestClusterPartitionAssignment(t *testing.T) {
```
jerry-stripe: This only covers a very basic case. I'm currently trying to test the flow of:
1. spin up 5 nodes, ensure they all have expected partitions
2. kill a few nodes, ensure that existing nodes remain
3. spin up nodes to replace old nodes, since they won't have same hostnames, ensure that they have partitions based on new ordering
4. make new version available, ensure all partitions become uniformly distributed again

jerry-stripe: Oh disregard this, I'm totally mistaken

vasi-stripe:
Just wanted to note our recent chat about ripping out Shard IDs.

It's concerning because if a node goes down, we have no way of exactly replacing it. This means if any three nodes go down (with replication = 3), we start serving 404s, even if we've added new nodes. We should at least have a plan for what to do if nodes start going down.

This also potentially makes any rebalancing operation more complex, because the assignment may be constantly changing as nodes are added/removed.

jerry-stripe:
We decided to keep Shard IDs around so closing in favor of https://github.com/stripe/sequins/pull/128

PR #126, opened 2017-10-23
initialize dbStatus with Version to prevent status page 502

jerry-stripe:
Fixes the case where the `Versions` map inside the `dbStatus` struct was referred to when uninitialized. Will provide code path through comments to be more clear about how this was happening. I'm not 100% confident that this was _the_ problem, but I think this is definitely a problem.

r? @scottjab-stripe 
cc? @stripe/storage

### DIFF
--- a/status.go
+++ b/status.go
@@ -69,7 +69,7 @@ type status struct {
 }
 
 type dbStatus struct {
-	Versions map[string]versionStatus `json:"versions,omitempty"`
+	Versions map[string]versionStatus `json:"versions"`
 }
 
 type versionStatus struct {
@@ -283,7 +283,7 @@ func (s *sequins) getPeerStatus(peer string, db string) (status, error) {
 	if db == "" {
 		err = decoder.Decode(&status)
 	} else {
-		s := dbStatus{}
+		s := dbStatus{Versions: make(map[string]versionStatus)}
 		err = decoder.Decode(&s)
 		if err != nil {
 			return status, err


jerry-stripe:
- https://github.com/stripe/sequins/blob/master/status.go#L177: `peerStatus` is initialized
- https://github.com/stripe/sequins/blob/master/status.go#L286: `dbStatus` is left empty, returning `peerStatus` with the form of `{DBs: map[string]dbStatus}`.
- https://github.com/stripe/sequins/blob/master/status.go#L186: the empty `dbStatus` struct is passed into the `mergeDBStatus` function as `left`
- https://github.com/stripe/sequins/blob/master/status.go#L303: exception is thrown by trying to assign into `Versions` map which is `nil`; surprisingly, even though the map is nil, the line above won't throw an exception because the default value is just returned with `ok` being `false`

bartle-stripe:
LGTM

Good find! I was confused how remotes could return return statuses with missing `Versions` until I realized we have `omitempty` listed on https://github.com/stripe/sequins/blob/master/status.go#L72.

bartle-stripe:
We might actually just want to remove that `omitempty` too, since there might be other things parsing the status output that aren't guarding against an empty `Versions`.

jerry-stripe:
Ohh good point, removed `omitempty`, PTAL @bartle-stripe

bartle-stripe:
lgtm

PR #127, opened 2017-10-26
Shard IDs: Don't wait for convergence too much

vasi-stripe:
The current convergence pattern isn't great. We wait to converge *twice* while holding a lock. And the lock prevents other nodes from joining. This causes a new cluster to converge on too-small a state while most nodes are blocked. This made it hard to write tests.

We can take the second wait to converge out of the lock entirely. And the first one we can shorten, since while the lock is held we can be pretty sure nodes aren't joining.

### DIFF
--- a/sequins.go
+++ b/sequins.go
@@ -28,6 +28,14 @@ import (
 
 const defaultMaxLoads = 10
 
+// While we wait to pick a shard ID, other nodes are prohibited from joining the cluster because
+// we're holding a lock. This timeout can therefore be pretty short.
+//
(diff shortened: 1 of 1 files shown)


scottjab-stripe:
LGTM

vasi-stripe:
I ran `make tests_functional` a number of times, and it failed once :(  Will investigate.

vasi-stripe:
I'm seeing similar rates of failures on master (~10% of test_functional runs have a failure). That still sucks, but I think this can go through.

PR #128, opened 2017-10-26
replace hash ring partition assignment with sequential scheme

jerry-stripe:
**Summary**
Changes the partition assignment scheme to use a sequential assignment to nodes as opposed to a hash ring. Essentially the same as https://github.com/stripe/sequins/pull/124, except keeping the Shard ID intact which is used to order nodes.

**Motivation**
Will distribute partitions amongst nodes much more uniformly.

r? @scottjab-stripe 
cc? @stripe/storage 

**Note**: Tests for this will not pass until https://github.com/stripe/sequins/pull/127 is merged in as we discovered a bug in the Shard ID logic that caused tests added in this PR to fail. Leaving as WIP until it gets merged, but this has already been tested to work assuming those changes.

### DIFF
--- a/cluster_test.go
+++ b/cluster_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"math/rand"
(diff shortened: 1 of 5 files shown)


jerry-stripe:
Rebased on top of #127, PTAL @scottjab-stripe

scottjab-stripe:
LGTM

God speed sir.

PR #130, opened 2017-10-30
Report the Shard ID on the status page

vasi-stripe:
This will let us add alerts for duplicates.

r? @jerry-stripe

### DIFF
--- a/status.go
+++ b/status.go
@@ -65,7 +65,8 @@ func init() {
 }
 
 type status struct {
-	DBs map[string]dbStatus `json:"dbs"`
+	DBs     map[string]dbStatus `json:"dbs"`
+	ShardID string              `json:"shard_id"`
 }
 
 type dbStatus struct {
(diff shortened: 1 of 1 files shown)


debug.go
```diff
@@ -67,7 +69,7 @@ func startDebugServer(config sequinsConfig) {
 
 	if config.Debug.Expvars {
 		mux.HandleFunc("/debug/vars", expvarHandler)
-		expStats = newStats(config.LocalStore)
+		expStats = newStats(config.LocalStore, sq.peers.ShardID)
```
jerry-stripe: Does this not cause problems when sequins is run locally? I think `sq.peers` can be `nil` if sharding isn't on so we might have to guard against that case.

vasi-stripe:
Now on the status page, not the debug page. And now checks if sharding is enabled first—nice catch there.

I also added back some older replication measures in the status JSON, because our Datadog checks were crashing, see https://github.com/stripe/datadog-checks/blob/a7a9e0f3b878a0ffcfba65f848473843fd196c13/checks.d/sequins.py#L34-L36

scottjab-stripe:
LGTM

PR #131, opened 2017-10-31
Log requests

vasi-stripe:
This will let us play them back later against multiple sequins clusters. Then when we make major changes to sequins, we can verify that everything is working just as before.


scottjab-stripe:
LGTM

PR #132, opened 2017-11-01
Allow faster startup, by deleting ZK nodes in parallel

vasi-stripe:
Sometimes it takes minutes for this to run.

### DIFF
--- a/zk/watcher.go
+++ b/zk/watcher.go
@@ -428,9 +428,15 @@ func (w *Watcher) cleanupTree(node string) {
 		return
 	}
 
+	var wg sync.WaitGroup
+	wg.Add(len(children))
 	for _, child := range children {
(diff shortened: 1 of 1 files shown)


jerry-stripe:
I'm so down for speeding up boot time, though https://github.com/stripe/sequins/pull/132/files#diff-085ac95e7286e4e0b4f7f05d0381ac81R416 makes it seem like `cleanupTree` isn't thread safe since it takes out a lock.

vasi-stripe:
I'm pretty sure that it's thread-safe, though. If you look at what `Delete()` actually does, it just calls `request()`, which enqueues something.

I'm not sure what we're locking to protect against—maybe so we don't delete something at the same time we try to watch it, or something?

jerry-stripe:
In that case, lgtm!

PR #134, opened 2017-11-02
remove use of versionAvailable, clarify documentation

jerry-stripe:
**Summary**
Remove the use of the `AVAILABLE` state and have the `BUILDING` state persist until `ACTIVE`. Make documentation a bit more clear as to what each state means.

**Motivation**
There has been a bit of confusion as to what each state means. Hopefully this clears some of it up.

**Testing**
Deployed to some smaller QA clusters so far, will work way up to rest of QA if all goes well.

r? @scottjab-stripe 
cc? @stripe/storage

### DIFF
--- a/db.go
+++ b/db.go
@@ -141,7 +141,6 @@ func (db *db) refresh() error {
 func (db *db) switchVersion(version *version) bool {
 	// Prepare the version, so that during the switching period we can respond
 	// to requests for it.
-	version.setState(versionAvailable)
 	db.mux.prepare(version)
 
 	// Build any partitions we're missing in the background.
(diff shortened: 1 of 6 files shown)


jerry-stripe:
I've also noticed that occasionally, a cluster will get stuck in which all versions were marked as `AVAILABLE` even though all of the data was done being downloaded. I don't believe that this will fix that problem, as they will just all be stuck in `BUILDING` but at least it will be a bit more accurate. 

I've also noticed that it was possible for some nodes in the cluster to be marked as `ACTIVE` while the rest were still `AVAILABLE`. My understanding was that this shouldn't really be possible so I changed the docs to reflect that. We'll definitely want to track down that issue in this or another PR I think.

jerry-stripe:
Since @scottjab-stripe is PTO, PTAL @vasi-stripe

vasi-stripe:
lgtm. As always, test in QA :)

PR #135, opened 2017-11-03
If we have a manifest, don't recalculate what to fetch

vasi-stripe:


scottjab-stripe:
LGTM

Maybe deploy this branch to QA first before merging.

vasi-stripe:
Confirmed on QA balancessequins that this does what we want.

Will now deploy to QA asgdecibelsequins.

PR #136, opened 2017-11-03
Revert "Shard IDs: Don't wait for convergence too much"

jerry-stripe:
This reverts commit 8d57ee513bb6a470c9ba453d3a27f8c032371fe3.

For whatever reason, it seems to be causing issues when rolling a cluster pre-this change. Such that the nodes being rolled will enter a new zk cluster and thus fill up on disk since they see no peers.

r? @vasi-stripe 
cc? @stripe/storage

### DIFF
--- a/sequins.go
+++ b/sequins.go
@@ -28,14 +28,6 @@ import (
 
 const defaultMaxLoads = 10
 
-// While we wait to pick a shard ID, other nodes are prohibited from joining the cluster because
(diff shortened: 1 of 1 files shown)


jerry-stripe:
Here I am wondering why the same tests keeps failing then I remembered that the convergence changes were supposed to fix that :/

I'm gonna look into a way to make them work without.

PR #137, opened 2017-11-07
Display shard IDs on status page

vasi-stripe:
* Show the shard ID for each node
* Show the nodes in shard-ID order

This should make things like duplicate shard IDs easier to notice.

image

r? @jerry-stripe


jerry-stripe:
Awesome! lgtm

PR #138, opened 2017-11-07
dedup shard IDs when picking local partitions

jerry-stripe:
**Summary**
Only consider unique shard IDs when choosing partition assignments.

**Motivation**
This will allow nodes to share partition assignments by using the same shard ID such that one node can replace another in the case of failure.

r? @vasi-stripe 
cc? @stripe/storage

### DIFF
--- a/doc/manual/2-2-sharding/README.md
+++ b/doc/manual/2-2-sharding/README.md
@@ -95,8 +95,9 @@ backfills them. To do that, it
     a. Creating an array of all partitions in sorted order * the replication
        factor. (e.g. [1, 1, 1, 2, 2, 2, ...])
 
-    b. Creating an array of all nodes it knows about in sorted order based on
-       their shard id.
+    b. Creating an array of all shard ids it knows about in sorted order. If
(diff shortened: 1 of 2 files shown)


vasi-stripe:
This looks good! Let's test it out in QA before we merge :)

status.go
```diff
@@ -432,7 +432,6 @@ func (vs *version) status() versionStatus {
 		CreatedAt:  vs.created.UTC().Truncate(time.Second),
 		State:      vs.state,
 		Partitions: partitions,
-		ShardID:    vs.sequins.config.Sharding.ShardID,
```
jerry-stripe: Probably should have put this in another PR, but I forgot to flag this in yours 😅 

This misses a lot of cases where the ShardID can be set (self-assigned id file, smallest missing, etc) and the Shard ID is already exposed in the header of requests so I feel like this isn't needed. Unless you explicitly wanted to expose what was set in the config file?

jerry-stripe:
@vasi-stripe Deployed to standardsequins in QA and everything looks peachy 👍

vasi-stripe:
lgtm

PR #139, opened 2017-11-07
fix exposure of shard id in status page

jerry-stripe:
Completely missed this in https://github.com/stripe/sequins/pull/137, but this just fixes the way we get the ShardID since it can be set by the self_assigned_id file or assigned by smallest misssing.

r? @vasi-stripe 
cc? @stripe/storage


vasi-stripe:
lgtm

PR #140, opened 2017-11-08
update zk url as the version we use got moved to archives

jerry-stripe:
**Summary**
Update the zk url, the zk version we use got moved to the archives.

**Motivation**
Travis builds have been failing with this problem: https://travis-ci.org/stripe/sequins/builds/299245524#L484

r? @vasi-stripe 
cc? @stripe/storage


vasi-stripe:
lgtm

PR #141, opened 2017-11-10
Small UI fixes.

jbalogh-stripe:
* give dbs a fixed width so each section is consistent and scannable
* drop bold from version names so that every word on the page is not bold

r? @jerry-stripe

### DIFF
--- a/status.tmpl
+++ b/status.tmpl
@@ -57,7 +57,7 @@
       }
 
       div.db {
-        min-width: 800px;
+        width: 800px;
         padding: 20px;
         margin: 0 auto 20px auto;
         display: table;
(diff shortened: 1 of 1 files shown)


jerry-stripe:
lgtm!

bartle-stripe:
I think this caused the partition table to wrap, making it unreadable :( I suspect what we want is to just use ~100% of the screen width?

jerry-stripe:
@bartle-stripe Which cluster are you seeing this on? This change hasn't been rolled out anywhere yet. I will try playing around with 100% screen width to see if it doesn't make things too hard to read.

bartle-stripe:
I just noticed it while debugging a single node (that had a custom build deployed to it).

jerry-stripe:
Oh yeah I see it now, will increase the width considerably to compensate

PR #142, opened 2017-11-14
add endpoint to allow sequins to fetch additional partitions for a gi…

jerry-stripe:
**Summary**
Add the `/fetch` endpoint (only accessible from `localhost`) that can be used to tell sequins to fetch the given partitions by sending a POST request with a JSON payload of the given form:
```
{
    "db1": {
        "version1": [1,2,3,4,5,6],
        "version2": [1,2,3,4,5,6]
    },

    "db2": {
        "version1": [1,2,3,4,5,6]
    }
}
```
If the request fails in any way, an InternalServerError (500) is returned, otherwise a 200 is returned. Additionally, nonexistent db/version pairs will be ignored rather than causing an error.

This also fixes the fact that the getters for the selected partitions would return the reference to the internal map ([here](https://github.com/stripe/sequins/compare/jerry-fetch?expand=1#diff-76451cb05f1dcf66ac577e04ec386a7dR147)). And the fact that this map was used in a way that assumed all of the values were `true` ([example](https://github.com/stripe/sequins/compare/jerry-fetch?expand=1#diff-493095f7c1f1cc2111c53a48c8823a99L130))

**Motivation**
This endpoint will allow us to implement a way to tell a sequins cluster to pick up under-replicated partitions.

**Concerns**
I've noticed the distinction between selected partitions and local partitions. I'm not sure if this is still a distinction that we feel strongly about but it causes the following behavior: 

Partitions grabbed through this method will persist between reboots of sequins and thus continue to take up disk space, despite the fact that they won't stay "selected". This is because upon starting up, sequins will pick the selected partitions based on its partition assignment, regardless of what it currently has. We might want to either, let sequins serve everything it has on disk (which is probably not a good idea as nodes will take a lot of traffic instead of proxying), or keep track of partitions fetched through this method and clean them up upon shutdown.

**Testing**
Tested against local sequins cluster and dummysequins.

r? @vasi-stripe, @scottjab-stripe 
cc? @stripe/storage

### DIFF
--- a/build.go
+++ b/build.go
@@ -50,13 +50,13 @@ func compressionCodecString(c sequencefile.CompressionCodec) string {
 	}
 }
 
-func (vs *version) build() {
+func (vs *version) build(rebuild bool) {
 	// Welcome to the sequins museum of lock acquisition. First, we grab the lock
 	// for this version, and check that the previous holder didn't finish
 	// building.
 	vs.buildLock.Lock()
 	defer vs.buildLock.Unlock()
-	if vs.built {
+	if !rebuild && vs.built {
 		return
 	}
 
@@ -159,7 +159,7 @@ func (vs *version) addFiles(partitions map[int]bool) error {
 	case err := <-errs:
 		return err
 	case <-c:
-		return vs.blockStore.Save(vs.partitions.SelectedLocal())
+		return vs.blockStore.Save(vs.partitions.Selected())
 	}
 }
 
--- a/sequins.go
+++ b/sequins.go
@@ -28,6 +28,8 @@ import (
 
 const defaultMaxLoads = 10
 
+const localhost = "127.0.0.1"
+
 // While we wait to pick a shard ID, other nodes are prohibited from joining the cluster because
 // we're holding a lock. This timeout can therefore be pretty short.
 //
@@ -375,6 +377,11 @@ func (s *sequins) refreshAll() {
 }
 
 func (s *sequins) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.URL.Path == "/fetch" && strings.HasPrefix(r.RemoteAddr, localhost) && r.Method == "POST" {
+		s.serveFetch(w, r)
+		return
+	}
+
 	if r.Method != "GET" {
 		w.WriteHeader(http.StatusBadRequest)
 		return
--- a/blocks/block_store.go
+++ b/blocks/block_store.go
@@ -105,7 +105,7 @@ func (store *BlockStore) Add(key, value []byte) error {
 
 // Save saves flushes any newly created blocks, and writes a manifest file to
 // the directory.
-func (store *BlockStore) Save(selectedPartitions map[int]bool) error {
+func (store *BlockStore) Save(selectedPartitions []int) error {
 	store.blockMapLock.Lock()
 	defer store.blockMapLock.Unlock()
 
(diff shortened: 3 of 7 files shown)


vasi:
I'll look at this in more detail tomorrow. Some notes:

* Is it safe comparing r.RemoteAddr to localhost? I think I'd feel comfier just listening on localhost only for /fetch
* We need to think much more carefully about locking. Is it safe to call `version.partitions.HaveSelected(p)` at the point we do? What if the version is currently being removed?
** Also, what happens if we call `version.build(true)` while there's another build going on, or when we have full download queue? Do we just wait for O(minutes/hours) before returning a status?
* Do we have any way to verify that the input is well-partitioned before attempting a fetch? If it's _not_ well-partitioned (eg: if someone runs this against standardsequins), we'll end up re-downloading every single data source.
* Currently, sequins should be picking the partitions to serve based on what's on-disk, rather than recalculating. I think it's best to maintain this behaviour, since otherwise the state of the cluster will change on rolling, which will be pretty confusing! Maybe a future command can clear overreplicated partitions to deal with this?

jerry-stripe:
@vasi No need to respond anytime soon, also responding with some notes :)
- Didn't know binding to localhost was an option, will definitely look into this.
- All access to the `selected` map is protected with a lock now, but I think you're right in that I need to look at this more, especially with regards to version removal.
   ** This endpoint will actually just launch `version.build` in a goroutine and return, similar to what happens with a normal version build. And since it takes the build lock, from my testing, running this multiple times doesn't cause issues because the builds will happen serially and then subsequent builds will see that the call to `NeededLocal` will yield nothing.
- To be honest I didn't consider such safe guards as I was imagining this to be just a low level way to tell sequins to grab data and mostly an interface that scripts could use. I don't know how hard it would be to determine if the input is well-partitioned, but it should be pretty straightforward to at least determine that we're not going to over-replicate some partitions so I'll definitely add that in.
- Based on the behavior that I've observed and [this function](https://github.com/stripe/sequins/blob/19defa8dadc227fbc6b2f10aeaf08e4cf752e808/sharding/partitions.go#L77-L117), it really seems like Sequins will just determined what should be served based on partition assignments and only serve those. Is there other code that determines what partitions are currently selected?

sequins.go
```diff
@@ -375,6 +377,11 @@ func (s *sequins) refreshAll() {
 }
 
 func (s *sequins) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.URL.Path == "/fetch" && strings.HasPrefix(r.RemoteAddr, localhost) && r.Method == "POST" {
```
scottjab-stripe: What happens if a db is named `fetch` Maybe we should update the docs to reflect that fetch and healthz are reserved db names.

build.go
```diff
@@ -50,13 +50,13 @@ func compressionCodecString(c sequencefile.CompressionCodec) string {
 	}
 }
 
-func (vs *version) build() {
+func (vs *version) build(rebuild bool) {
```
dustyburwell: Total nit, and I don’t know about the idiomacy of this in golang, but I dislike Boolean flags like this preferring an explicit second function with reuse.

sequins.go
jerry-stripe: It has to be a post request to trigger the request, so I think this is safe.

build.go
jerry-stripe: Yeah I think you're right here, I'll update with reuse.

jerry-stripe:
Closing this in favor of another method

PR #143, opened 2017-11-14
update replication status for versionStatus struct on initialization

jerry-stripe:
**Summary**
Updates the replication information for a `versionStatus` on creation and allows it to be updated by calling `calculateReplicationStats` multiple times. Previously, we would only set it after collecting all of our peers which would be a problem when getting the status with `?proxy=no`.

**Motivation** 
Ensure that replication info is up to date regardless of how it is accessed.

r? @bartle-stripe 
cc? @stripe/storage

### DIFF
--- a/status.go
+++ b/status.go
@@ -74,10 +74,12 @@ type dbStatus struct {
 }
 
 type versionStatus struct {
-	Path                 string      `json:"path"`
-	NumPartitions        int         `json:"num_partitions"`
+	Path              string `json:"path"`
+	NumPartitions     int    `json:"num_partitions"`
+	TargetReplication int    `json:"target_replication"`
+
+	// Values that are recalculated with calculateReplicationStats
 	ReplicationHistogram map[int]int `json:"replication_histogram"`
-	TargetReplication    int         `json:"target_replication"`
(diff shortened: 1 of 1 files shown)


status.go
```diff
@@ -201,8 +203,10 @@ func (s *sequins) serveStatus(w http.ResponseWriter, r *http.Request) {
 		}
 
 		for _, db := range status.DBs {
-			for versionName, version := range db.Versions {
-				db.Versions[versionName] = calculateReplicationStats(version)
+			for versionName := range db.Versions {
+				vst := db.Versions[versionName]
+				vst.calculateReplicationStats()
+				db.Versions[versionName] = vst
```
bartle-stripe: Why do we need this assignment?

jerry-stripe: @bartle-stripe  I can't modify it in place because apparently you can't take the address of map values in go. The alternative was to make the function not take a receiver and just reassign it, but that didn't really seem any cleaner.

bartle-stripe: I guess you could change `db.Versions` to `map[string]*versionStatus`, but meh.

bartle-stripe:
lgtm

PR #144, opened 2017-11-14
increase css width some more

jerry-stripe:
Increase the width of the expanded db view to better accommodate the partition map.

r? @bartle-stripe 
cc? @stripe/storage


bartle-stripe:
lgtm

PR #145, opened 2017-11-16
add max_replication to config to allow for a partition replication limit

jerry-stripe:
**Summary**
Add a new config option, `max_replication` that will limit the replication factor of partitions. For a given node, when choosing which partitions it should have locally, it will first ensure that selecting a partition wouldn't bring it over this value. If it would, it doesn't select it. `max_replication` is only respected if `max_replication <= replication`, so setting it to 0 will disable this feature.

**Motivation**
This will limit the damage of ZK flapping such that nodes don't fill up too rapidly.

**Testing**
Unit testing for now. Will test in QA as well.

r? @scottjab-stripe 
cc? @stripe/storage

### DIFF
--- a/cluster_test.go
+++ b/cluster_test.go
@@ -109,9 +109,21 @@ func minRepl(r int) configOption {
 	}
 }
 
+func maxRepl(r int) configOption {
+	return func(config *sequinsConfig) {
+		config.Sharding.MaxReplication = r
+	}
+}
+
 func noShardID() configOption {
 	return func(config *sequinsConfig) {
 		config.Sharding.ShardID = ""
+	}
+}
+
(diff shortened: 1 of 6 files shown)


jerry-stripe:
Tested following cases in QA and seemed to respond appropriately 👍 
1. spin up cluster with 10 nodes, wait for them to all fetch
2. add node, with `max_replication = replication`, no fetching as expected
3. add node, with `max_replication = replication + 1`, fetched as normal
4. add node with `max_replication = 0`, fetched as normal
5. add node with `max_replication = replication` no fetching as expected

scottjab-stripe:
LGTM

PR #146, opened 2017-11-16
emit an event when a version is set to active in a db

jerry-stripe:
**Summary**
Emit a metric when a version is set to active in a db.

**Motivation**
This is a feature that has been asked for an is probably a useful thing to emit.

r? @scottjab-stripe 
cc? @stripe/storage

### DIFF
--- a/db.go
+++ b/db.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"net/http"
@@ -9,6 +10,8 @@ import (
 	"path/filepath"
 	"sync"
 	"time"
+
+	"github.com/DataDog/datadog-go/statsd"
 )
 
 var errNoVersions = errors.New("no versions available")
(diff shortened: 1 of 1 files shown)


db.go
```diff
@@ -203,6 +206,14 @@ func (db *db) upgrade(version *version) {
 	db.mux.upgrade(version)
 	version.setState(versionActive)
 
+	if version.stats != nil {
+		title := fmt.Sprintf("A new version is active in %s.", db.name)
+		text := fmt.Sprintf("Version %s has been set to active in db %s.", version.name, db.name)
+		event := statsd.NewEvent(title, text)
```
scottjab-stripe: Does this honor the config to override the statsd url?

jerry-stripe: It should, the url is set when [the client is created](https://github.com/stripe/sequins/blob/56fc7e96de1dee46ce50a3d70aedb7680b96c102/sequins.go#L73-L81). I'll also just manually confirmed

jerry-stripe: Whoops meant to say that I did manually confirm.

scottjab-stripe:
LGTM

PR #147, opened 2017-11-16
lock access to p.selected

jerry-stripe:
**Summary**
Protect access to `p.selected` with a lock, and also have the wrapper function return a alice since both consumers convert to a slice anyways.

r? @scottjab-stripe 
cc? @stripe/storage

### DIFF
--- a/blocks/block_store.go
+++ b/blocks/block_store.go
@@ -105,7 +105,7 @@ func (store *BlockStore) Add(key, value []byte) error {
 
 // Save saves flushes any newly created blocks, and writes a manifest file to
 // the directory.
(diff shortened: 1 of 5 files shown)


scottjab-stripe:
The code looks fine, but what is the motivation behind protecting p.selected?

jerry-stripe:
Mostly just slightly paranoid about later running into issues with it. I think I was starting to interact with it more in https://github.com/stripe/sequins/pull/142 and just wanted to lock it down to make sure I didn't introduce races. Since we're going down another route, this probably isn't necessary so we can close if it's not worth it.

PR #148, opened 2017-11-22
allow s3_backend to handle _SYMLINK

jerry-stripe:
**Summary**
Add the ability to handle s3 versions symlinks through the `_SYMLINK` file. When sequins encounters a version containing a `_SYMLINK` file, it will redirect all file Open and ListFiles requests to the target version as determined by what's in the `_SYMLINK` file.

**Motivation**
This will allow us to force sequins to refetch a version by creating a new version with a lexicographically greater name and a `_SYMLINK` file with the target version specified. This is a much faster method than copying S3 directories.

**Testing**
Tested against QA.

r? @scottjab-stripe 
cc? @stripe/storage

### DIFF
--- a/backend/s3_backend.go
+++ b/backend/s3_backend.go
@@ -3,6 +3,7 @@ package backend
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"path"
 	"sort"
@@ -45,9 +46,9 @@ func (s *S3Backend) ListVersions(db, after string, checkForSuccess bool) ([]stri
 		var filtered []string
 		for _, version := range versions {
 			successFile := path.Join(s.path, db, version, "_SUCCESS")
-			exists := s.exists(successFile)
+			symlinkFile := path.Join(s.path, db, version, "_SYMLINK")
 
-			if exists {
+			if s.exists(successFile) || s.exists(symlinkFile) {
 				filtered = append(filtered, version)
 			}
 		}
@@ -122,7 +123,13 @@ func (s *S3Backend) listDirs(dir, after string) ([]string, error) {
 }
 
 func (s *S3Backend) ListFiles(db, version string) ([]string, error) {
-	versionPrefix := path.Join(s.path, db, version)
+	// If this directory is a symlink (denoted with the _SYMLINK file), we
+	// should return the contents of the directory it is linking to.
+	targetVersion, err := s.getTargetVersion(db, version)
+	if err != nil {
+		return nil, err
+	}
+	versionPrefix := path.Join(s.path, db, targetVersion)
 
 	// We use a set here because S3 sometimes returns duplicate keys.
 	res := make(map[string]bool)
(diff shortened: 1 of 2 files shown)


s3_sequins_test.go
```diff
@@ -21,21 +22,20 @@ import (
 
 type S3ClientMock struct {
 	s3iface.S3API
-	toggle             bool
+	counter            int
```
jerry-stripe: Had to change the way this test worked since `GetObjectOutput` is now called more frequently due to checks for the `_SYMLINK` file.

backend/s3_backend.go
```diff
@@ -156,7 +163,6 @@ func (s *S3Backend) ListFiles(db, version string) ([]string, error) {
 
 	log.Printf("call_site=s3.ListFiles sequins_db=%q sequins_db_version=%q dataset_size=%d file_count=%d", db, version, datasetSize, numFiles)
```
jerry-stripe: I'm a bit back and forth on whether when logging, we should log the original version or the target version 😕

backend/s3_backend.go
```diff
@@ -196,6 +202,40 @@ func (s *S3Backend) Open(db, version, file string) (io.ReadCloser, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return s.open(db, targetVersion, file)
+}
+
+// getTargetVersion returns the target version of a symlinked version if the
+// _SYMLINK file exists or the same version otherwise.
+func (s *S3Backend) getTargetVersion(db, version string) (string, error) {
```
jerry-stripe: I feel like since this gets called for every `Open`, is it worth it to cache results?

jerry-stripe:
PR is ready now, PTAL @scottjab-stripe :)

backend/s3_backend.go
scottjab-stripe: Log that we are using a symlink, db version should be based from current s3 path where sym link file is, and the symlinked_version would be the OG version.

backend/s3_backend.go
scottjab-stripe: Maybe have an option to skip this check? If we are just downloading files to load a version, no point in grabbing this file.

backend/s3_backend.go
```diff
@@ -3,6 +3,7 @@ package backend
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
```
scottjab-stripe: go imports?

scottjab-stripe:
Except for the few nits, I like this. 

I don't think this would happen, but what happens when you have symlinked versions that point to each other?

backend/s3_backend.go
jerry-stripe: I was thinking since the `Open` function is public and the caller has no knowledge of the fact that a version may be backed by a symlinked directory, we need to check with each call.

backend/s3_backend.go
jerry-stripe: It's used here: https://github.com/stripe/sequins/pull/148/files#diff-46f6ba2d612aec109d60649c7f7a9268R231, I don't think it would have compiled otherwise

jerry-stripe:
@scottjab To answer your question, from my testing, the s3_backend will attempt to list all files, see nothing and the version will come online with no files. The script used to create symlinks will copy the contents of the `_SYMLINK` file if it exists so this _shouldn't_ happen. It's definitely a case that's fixable, but doesn't seem worth it to handle imo.

scottjab-stripe:
That sounds fine to me

jerry-stripe:
PTAL @scottjab-stripe

scottjab-stripe:
LGTM

PR #149, opened 2017-11-29
emit a metric instead of an event when version is active

jerry-stripe:
Switch to using a metric instead of an event so that they can get piped into redshift.

r? @scottjab-stripe 
cc? @stripe/storage

### DIFF
--- a/db.go
+++ b/db.go
@@ -10,8 +10,6 @@ import (
 	"path/filepath"
 	"sync"
 	"time"
-
-	"github.com/DataDog/datadog-go/statsd"
 )
 
 var errNoVersions = errors.New("no versions available")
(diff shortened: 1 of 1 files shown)


db.go
```diff
@@ -207,11 +205,9 @@ func (db *db) upgrade(version *version) {
 	version.setState(versionActive)
 
 	if version.stats != nil {
-		title := fmt.Sprintf("A new version is active in %s.", db.name)
-		text := fmt.Sprintf("Version %s has been set to active in db %s.", version.name, db.name)
-		event := statsd.NewEvent(title, text)
-		event.Tags = []string{fmt.Sprintf("sequins_db:%s", db.name)}
-		version.stats.Event(event)
+		duration := version.active.Sub(version.created)
+		tags := []string{fmt.Sprintf("sequins_db:%s", db.name), fmt.Sprintf("version:%s", version.name)}
```
jerry-stripe: @scottjab-stripe Do you think that by including the version, the cardinality is too high for datadog?

scottjab-stripe: I do, we got turned off about a year ago when I tried this. We do emit a log line when a version becomes active, maybe we could emit more stats that are parseable when that happens.

jerry-stripe:
Turns out we are already doing this with the `sequins.db_creation_time` metric.

PR #150, opened 2018-02-07
strip contents of _SYMLINK file

jerry-stripe:
Trim the contents of the `_SYMLINK` file to buffer against potential whitespace when manually creating the file.

r? @vasi-stripe 
cc? @stripe/storage

### DIFF
--- a/backend/s3_backend.go
+++ b/backend/s3_backend.go
@@ -48,8 +48,20 @@ func (s *S3Backend) ListVersions(db, after string, checkForSuccess bool) ([]stri
 			successFile := path.Join(s.path, db, version, "_SUCCESS")
 			symlinkFile := path.Join(s.path, db, version, "_SYMLINK")
 
-			if s.exists(successFile) || s.exists(symlinkFile) {
+			if s.exists(successFile) {
 				filtered = append(filtered, version)
+			} else if s.exists(symlinkFile) {
+				// For symlinked versions, we check that the
+				// target version has a success file.
+				targetVersion, err := s.getTargetVersion(db, version)
+				if err != nil {
+					continue
+				}
+
+				targetSuccessFile := path.Join(s.path, db, targetVersion, "_SUCCESS")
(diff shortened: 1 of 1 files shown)


vasi-stripe:
`Received unexpected error "InvalidAccessKeyId: The AWS Access Key Id you provided does not exist in our records.\n\tstatus code: 403, request id: DF2D7998F75CD416, host id: K3Kk2dnJp1af2brK3pZrZeVkepHaehT7459k9nWu1W6dq4p4q02s3C3XU8dw5/Pv0thSizuNzjo="`
😲

vasi-stripe:
Also, I think the more important change may be to check that the target directory actually exists and has a _SUCCESS file, before returning it from listDBs :)

jerry-stripe:
@vasi-stripe I'm not sure what's up with these tests :/
They fail at the same point when run locally as well:
```
--- FAIL: TestS3Backend (0.46s)
        Error Trace:    s3_sequins_test.go:53
                        s3_sequins_test.go:81
        Error:          Received unexpected error "BucketRegionError: incorrect region, the bucket is not in 'us-west-2' region\n\tstatus code: 301, request id: , host id: "
        Messages:       should be able to delete old test files on s3
--- FAIL: TestS3Sequins (0.18s)
        Error Trace:    s3_sequins_test.go:53
                        s3_sequins_test.go:74
                        s3_sequins_test.go:170
        Error:          Received unexpected error "BucketRegionError: incorrect region, the bucket is not in 'us-west-2' region\n\tstatus code: 301, request id: , host id: "
        Messages:       should be able to delete old test files on s3
```

I would make tests for this, but I think it would require non-trivial changes to s3 tests, and would rather invest the time in ng-sequins. Otherwise, I tested several cases locally, and am testing dummysequins now.

jerry-stripe:
yay, thanks for fixing the tests! PTAL @vasi-stripe

vasi-stripe:
lgtm

PR #151, opened 2018-02-09
Update go-zookeeper

vasi-stripe:
Should prevent recurring disconnects: https://github.com/samuel/go-zookeeper/pull/167


scottjab-stripe:
🙏 

LGTM

PR #152, opened 2018-02-09
Close once and delete once

ying-stripe:
Make sure (*version).close/delete is called only once. This prevents the race that causes sequins to panic when a version is being removed while the entire dataset is also being removed. In this case, close() may be called twice on different code paths. 

r? @scottjab-stripe @dusty-stripe

### DIFF
--- a/version.go
+++ b/version.go
@@ -48,6 +48,9 @@ type version struct {
 	buildLock sync.Mutex
 
 	stats *statsd.Client
+
+	closeOnce  sync.Once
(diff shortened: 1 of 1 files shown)


scottjab-stripe:
Tests are not passing, once they pass this looks sane.

ying-stripe:
@scottjab-stripe Tests passed.

scottjab-stripe:
LGTM

PR #153, opened 2018-02-12
Disable S3 tests

vasi-stripe:
We currently have some problems with our S3 test account, which won't get resolved for at least a few days, and are blocking some important PRs.


ying-stripe:
lgtm

PR #154, opened 2018-02-15
Ratelimit file downloads

ying-stripe:


sequins.go
```diff
@@ -100,6 +107,13 @@ func (s *sequins) init() error {
 	}
 	s.workQueue = workqueue.NewWorkQueue(maxLoads)
 
+	// Set up token bucket for download bandwidth throttling
+	maxDownloadBandwidth := s.config.MaxDownloadBandwidthMBPerSecond * 1024 * 1024
```
wei-stripe: Can we just not introduce this artificial limit and just pass nil to bucket?

In the downloader, if the bucket is nil, we can skip the reader wrap operation.

vendor/vendor.json
```diff
@@ -1659,6 +1665,12 @@
 			"path": "golang.org/x/net/lex/httplex",
 			"revision": "dd2d9a67c97da0afa00d5726e28086007a0acce5",
 			"revisionTime": "2017-02-23T02:43:42Z"
+		},
+		{
+			"checksumSHA1": "538I8ghUdvZa25NxgMUDpq14Qic=",
+			"path": "gopkg.in/check.v1",
```
vasi-stripe: Do we need to include this? It looks like juju/ratelimit only uses this for its tests, which we shouldn't be running, right?

vasi-stripe:
LGTM

vasi-stripe:
(modulo nitpicks)

sequins.go
ying-stripe: Okay. 

In my opinion, we should throttling by default. The downloading part has lower priority. It should not compete with the serving part on bandwidth and CPU.

vendor/vendor.json
ying-stripe: I used govendor to add juju/ratelimit. It added this automatically. If I remove it, next time when somebody runs govendor again, may it be added back? I will check it.

build.go
```diff
@@ -187,7 +198,7 @@ func (vs *version) sparkeyDownload(src, dst, fileType string, transform func(io.
 		}
 	}()
 
-	var trStream io.Reader = stream
+	var trStream = rateLimitedStream
```
wei-stripe: can be simplified to:
`trStream =: vs.rateLimitedDownloadStream(stream)`

wei-stripe:
lgtm

PR #156, opened 2018-02-15
Restore S3 tests

vasi-stripe:
Restore S3 tests, now that we have a good S3 test account again.

Also run tests verbosely, so that we can actually tell that S3 tests are running and not skipped.


scottjab-stripe:
God speed.

LGTM

PR #157, opened 2018-02-20
Die if sequins is ZK-flapping

vasi-stripe:
Our service runner should then restart sequins, allowing us to successfully reconnect. Yes, this is ugly :(

Also add some extra debugging info to our ZK watcher.

### DIFF
--- a/sequins.go
+++ b/sequins.go
@@ -39,6 +39,11 @@ const shardIDConvergenceMax = 3 * time.Second
 
 var errDirLocked = errors.New("failed to acquire lock")
 
+const (
+	flapMax      = 10
+	flapDuration = 5 * time.Minute
+)
+
 type sequins struct {
 	config  sequinsConfig
 	http    http.Handler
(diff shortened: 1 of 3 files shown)


sequins.go
```diff
@@ -216,6 +221,12 @@ func (s *sequins) initCluster() error {
 		return err
 	}
 
+	flapNotify := zkWatcher.SetFlapThreshold(flapMax, flapDuration)
+	go func() {
+		<-flapNotify
+		panic("Dying due to ZK flapping")
```
scottjab-stripe: I really don't like panics as they are hard to parse in splunk, can we change this to a `log.Fatal`.  We don't really care about the stack trace here do we?

zk/watcher_test.go
```diff
@@ -127,3 +130,46 @@ func TestZKRemoveWatch(t *testing.T) {
...
+
+	for i := 0; i < 5; i++ {
+		simulateError(w)
+	}
+	select {
+	case <-time.After(100 * time.Millisecond):
+		assert.Fail(t, "Flaps in a short period should notify")
+	case <-flapNotify:
+	}
+	println("DONE")
```
scottjab-stripe: Is this print needed?

vasi-stripe:
Fixed those nits. PTAL @scottjab-stripe

scottjab-stripe:
LGTM

PR #159, opened 2018-02-22
Use a feature flag to control fetching versions

vasi-stripe:
This lets Sequins start up in a mode where it only checks local disk for versions, rather than the network (mostly). It should enable rolling sequins with checking-for-new-versions off, and hence our feature flags use-case.

cc @scottjab-stripe @amith-stripe

### DIFF
--- a/cluster_test.go
+++ b/cluster_test.go
@@ -3,7 +3,6 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"math/rand"
 	"net"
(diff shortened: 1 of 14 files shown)


cluster_test.go
```diff
@@ -127,6 +127,13 @@ func shardID(id string) configOption {
 	}
 }
 
+func featureFlags(file *os.File) configOption {
```
amith-stripe: oh i need to remove this method, one sec

sequins.go
```diff
@@ -114,8 +129,15 @@ func (s *sequins) init() error {
 		s.downloadRateLimitBucket = ratelimit.NewBucketWithRate(float64(maxDownloadBandwidth), maxDownloadBandwidth)
 	}
 
+	// Kick off feature flag cache refresh if GoforitFlagJsonPath configured
+	if s.config.GoforitFlagJsonPath != "" {
+		log.Printf("Enabling Goforit: GoforitFlagJsonPath=%s", s.config.GoforitFlagJsonPath)
```
scottjab-stripe: use `%q` instead of `%s`  it will use the correct quotes around strings to make splunk happy.

sequins.go
```diff
@@ -326,10 +356,40 @@ func (s *sequins) listDBs() ([]string, error) {
...
 }
 
-func (s *sequins) refreshAll() {
+func (s *sequins) remoteRefresh() bool {
+	flagName := disableRemoteRefreshFlagPrefix
+	if s.config.Sharding.ClusterName != "" {
+		flagName += "." + s.config.Sharding.ClusterName
+	}
+	if s.goforit != nil && goforit.Enabled(context.Background(), flagName) {
+		log.Printf("Not allowing remote refresh: cluster=%s, flag=%s\n",
```
scottjab-stripe: `%q`  and you don't need `\n` with the logger.

vendor/github.com/stripe/goforit/flags.go
```diff
@@ -0,0 +1,243 @@
...
+
+func parseFlagsJSON(r io.Reader) (map[string]Flag, error) {
+	dec := json.NewDecoder(r)
+	var v JSONFormat
+	err := dec.Decode(&v)
+	if err != nil {
+		fmt.Printf("WTF? %s\n", err)
+		return nil, err
+	}
+	spew.Dump(v.Flags)
```
vasi-stripe: It looks like we accidentally vendored this in 😲 

Can you vendor in a clean master?

sequins.go
```diff
@@ -136,7 +158,15 @@ func (s *sequins) init() error {
 	signal.Notify(sighups, syscall.SIGHUP)
 	go func() {
 		for range sighups {
-			s.refreshAll()
+			// Refresh any feature flags on HUP
+			if s.goforit != nil {
+				err := goforit.RefreshFlags(s.goforit)
+				if err != nil {
+					log.Printf("Error refreshing flags on HUP: %s\n", err)
```
scottjab-stripe: `\n` is redundant with the logger.

db.go
```diff
@@ -45,16 +46,37 @@ func (db *db) listVersions(after string) ([]string, error) {
...
-func (db *db) backfillVersions() error {
+func (db *db) backfillVersions(initialLocal bool) error {
 	db.refreshLock.Lock()
 	defer db.refreshLock.Unlock()
 
-	versions, err := db.listVersions("")
+	var versions []string
+	var err error
+	if initialLocal {
+		log.Printf("Initial fetch of local versions only: db=%s\n", db.name)
```
scottjab-stripe: `\n`  (see other comments)

sequins.go
vasi-stripe: Oops I think that was my idea :(

vasi-stripe:
Looks like there's a data race coming up in tests, I can't repro locally. I'll take a look at that.

sequins.go
amith-stripe: ah cool!  i'll update the other ones i added in this PR

scottjab-stripe:
squash the commits, then I am good.

LGTM

vasi-stripe:
We'll need to re-vendor goforit after that gets merged, but that can happen later.

PR #161, opened 2018-02-23
Fix TriggerCleanup()

ying-stripe:
Serialize cleaning up znodes and do not clean up znodes when the db versions are still available in S3

r? @bartle-stripe  @vasi-stripe

### DIFF
--- a/sequins.go
+++ b/sequins.go
@@ -450,7 +450,7 @@ func (s *sequins) refreshAll(initialStartup bool) {
 
 	// Cleanup any zkNodes for deleted versions and dbs.
 	if s.zkWatcher != nil {
-		s.zkWatcher.TriggerCleanup()
+		go s.zkWatcher.TriggerCleanup()
 	}
 }
 
--- a/sequins.go
+++ b/sequins.go
@@ -257,8 +257,6 @@ func (s *sequins) initCluster() error {
 		log.Fatal("Dying due to ZK flapping")
 	}()
 
-	go zkWatcher.TriggerCleanup()
-
 	hostname := s.config.Sharding.AdvertisedHostname
 	if hostname == "" {
 		hostname, err = os.Hostname()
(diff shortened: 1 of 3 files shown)


vasi-stripe:
Hmm, I'm not sure about a couple of things here:

1. Do we want to completely serialize cleanupTree? Or just reduce the parallelism? I'd like an estimate of how long cleanupTree takes on a big tree (eg: standardsequins) when serialized.
2. I don't understand why it's useful to exclude things not in S3. By the time a version is removed from S3, it should have been dropped days ago.

ying-stripe:
1. If we exclude all db versions in S3, even if we serialize cleanupTree, it can be fast. I tested it on mlsequins. If there are no zk nodes to remove, it is about 0.2s. If we don't exclude db versions in S3, it will take much longer time for sure. 
2. The reason is that we don't want cleanupTree() to race with creating watches on any node. By the time a db version is removed from S3, it is unlikely any node is trying to create the path to the znode of the db version. So we can avoid the 'no does not exist' zookeeper errors that we see in sequin logs. The tradeoff is that we leave more empty znodes in zookeeper.

ying-stripe:
I tested it on standardsequins too. It is about 0.5s. I didn't count the extra S3 list calls outside TriggerCleanup().

vasi-stripe:
Ah, so we're simply not removing nodes at all if they're still in S3? Doesn't that tell other sequinses that we still have those partitions, even if we don't?

ying-stripe:
I don't understand the question. A sequins node tells other sequins node it owns a partition by creating a ephemeral znode under the znode for a db version. If it doesn't own a partition, it just removes the ephemeral znode not the znode for a db version. We can even run a script periodically to remove the empty non-ephemeral znodes. 

I am trying a simpler solution to avoid the race. Will let you know the result.

ying-stripe:
Close this one in favor of https://github.com/stripe/sequins/pull/163.

PR #162, opened 2018-02-27
Add more zk logging

ying-stripe:
- Add more zk logging
- Fix some issues zk/watcher.go
- Add a long pause detector

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"net/url"
 	"path/filepath"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
@@ -119,6 +120,24 @@ func main() {
 	if config.Debug.Bind != "" {
 		startDebugServer(config)
 	}
+
+	// This goroutine is supposed to wake up every 20 milliseconds. If it
+	// finds itself slept for an unusually long time, it writes a log entry
+	// to record the delay. The delay could be due to GC pause, scheduler
+	// issue or others. This may help us investigate the 504 issue that we
+	// have seen recently.
+	go func() {
+		lastWakeUpTime := time.Now()
+		for {
+			time.Sleep(20 * time.Millisecond)
+			now := time.Now()
+			delay := now.Sub(lastWakeUpTime) - 20*time.Millisecond
+			lastWakeUpTime = now
+			if delay > 20*time.Millisecond {
+				log.Printf("Detected %d milliseconds unusual delay", delay.Nanoseconds()/1000000)
+			}
+		}
+	}()
 
 	s.start()
 }
--- a/zk/watcher.go
+++ b/zk/watcher.go
@@ -93,7 +93,7 @@ func (w *Watcher) reconnect() error {
 	defer w.lock.Unlock()
 
 	servers := strings.Join(w.zkServers, ",")
-	log.Println("Connecting to zookeeper at ", servers)
+	log.Println("sequins ZK connecting to zookeeper at ", servers)
 	conn, events, err = zk.Connect(w.zkServers, w.sessionTimeout)
 	if err != nil {
 		return err
@@ -151,7 +151,7 @@ func (w *Watcher) runHooks() error {
 	w.hooksLock.Lock()
 	defer w.hooksLock.Unlock()
 
-	log.Println("Recreating ephemeral nodes")
+	log.Println("sequins ZK recreating ephemeral nodes")
 	for node := range w.ephemeralNodes {
 		err := w.createEphemeral(node)
 		if err != nil {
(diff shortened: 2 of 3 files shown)


zk/watcher.go
```diff
@@ -310,7 +322,7 @@ func (w *Watcher) createEphemeral(node string) error {
 		}
 	}
 
-	return nil
+	return fmt.Errorf("failed to create ephemeral znode %s after %d retries", node, maxCreateRetries)
```
bartle-stripe: Could we `err` from the final attempt?

ying-stripe: The final call is `err = w.createAll(parent)`. Only when err == nil, we will reach here.

bartle-stripe:
lgtm

zk/watcher.go
```diff
@@ -118,6 +124,9 @@ func (w *Watcher) reconnect() error {
 	}
 
 	go func() {
+		// QUESTION: why do we need to range over events? childrenW() also returns an event chanel.
```
scottjab-stripe: This is probably just because we ported the old c based library to the go native one about 2 years go.

zk/watcher.go
```diff
@@ -102,6 +104,10 @@ func (w *Watcher) reconnect() error {
 	}
 	w.conn = conn
 
+	// QUESTION: why do need to set connection timeout here?
```
scottjab-stripe: If I remember correctly the server disconnects were unreliable, so we started keeping track on the client side.

main.go
```diff
@@ -120,6 +121,24 @@ func main() {
 		startDebugServer(config)
 	}
 
+	// This goroutine is supposed to wake up every 20 milliseconds. If it
+	// finds itself slept for an unusually long time, it writes a log entry
+	// to record the delay. The delay could be due to GC pause, scheduler
+	// issue or others. This may help us investigate the 504 issue that we
+	// have seen recently.
+	go func() {
```
vasi-stripe: What would this find?

zk/watcher.go
```diff
@@ -67,6 +67,8 @@ func Connect(zkServers []string, prefix string, connectTimeout, sessionTimeout t
 		watchedNodes:   make(map[string]watchedNode),
 	}
 
+	// QUESTION: do we need to call reconnect here? I don't think it is necessary because we will call
+	// reconnect() inside w.run() in another goroutine immediately.
```
vasi-stripe: On the first loop inside `run`, we wait for errors.

vasi-stripe:
I don't know that I can answer most of your questions, I wasn't involved in writing this code :(

zk/watcher.go
```diff
@@ -159,6 +170,7 @@ func (w *Watcher) runHooks() error {
 	return nil
 }
 
+// QUESTION: Is this needed? I can't find anywhere wn.disconnected is read.
```
bartle-stripe: It's returned in `WatchChildren`, and `Peers.sync` later tries to read from it.

zk/watcher.go
```diff
@@ -142,13 +151,15 @@ func (w *Watcher) runHooks() error {
 	w.hooksLock.Lock()
 	defer w.hooksLock.Unlock()
 
+	log.Println("Recreating ephemeral nodes")
```
vasi-stripe: Let's add a common prefix (`sequins ZK`?) to all logs from this file.

vasi-stripe:
The errors we get in prod are from `sendErr("re-adding watch", w.errs, err, node, "")`

I have a sneaking suspicion that these are leftovers from the previous connection—is there a way we can test for that?

main.go
ying-stripe: If we find in the log that 'All peers timed out' happens together with this 'unusual delay', we can infer there may be some system-wide issue. It could be a long GC pause, CPU being throttled, scheduler issue...?

zk/watcher.go
ying-stripe: I feel we do unnecessary zookeeper connects. We can check the state of the connection. If it is already connected, we don't need to connect again and close the old connection.

zk/watcher.go
ying-stripe: Yes, you are right.

zk/watcher.go
ying-stripe: OK. I talked to the team a few days ago I am missing log4j.

zk/watcher.go
ying-stripe: The sample code implies that zk.Connect() is a synchronous call. When we reach here, we have already connected to zookeeper.

zk/watcher.go
ying-stripe: @scottjab-stripe We can clean up the code there. I checked  go-zookeeper. We will never get an event with event.Err != nil from the event channel returned by zk.Connect()

zk/watcher.go
ying-stripe: I debugged it a little. The first event we get is is an event with zk.StateConnecting. We get it immediately. It doesn't seem we have to wait the timeout here.

ying-stripe:
LGTM
(this has been tested in a QA sequins node)

PR #164, opened 2018-03-07
Update goforit in sequins

vasi-stripe:
Adds metrics

### DIFF
--- a/sequins.go
+++ b/sequins.go
@@ -160,12 +160,8 @@ func (s *sequins) init() error {
 		for range sighups {
 			// Refresh any feature flags on HUP
 			if s.goforit != nil {
-				err := goforit.RefreshFlags(s.goforit)
-				if err != nil {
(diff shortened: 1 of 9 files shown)


vasi-stripe:
r? @asoni-stripe 

Deployed to QA balancessequins

sequins.go
```diff
@@ -160,12 +160,8 @@ func (s *sequins) init() error {
 		for range sighups {
 			// Refresh any feature flags on HUP
 			if s.goforit != nil {
-				err := goforit.RefreshFlags(s.goforit)
-				if err != nil {
-					log.Printf("Error refreshing flags on HUP: %s", err)
```
asoni-stripe: ooc, why remove the log?

vasi-stripe: We no longer have an error return from RefreshFlags, that's all internal now :)

PR #165, opened 2018-03-07
Add request logging via feature flag

vasi-stripe:
We have request logging, but deploying it is a real pain in the butt on a large cluster.

Use a feature flag instead!


jerry-stripe:
LGTM

PR #166, opened 2018-03-12
Send DB request stats

vasi-stripe:
Send a count metric for every DB request. This helps to figure out what data sets are completely unused, and should probably be deleted.

* This only sends aggregate data every minute, so it's not a high load on sequins or DataDog
* In contrast to expvar metrics:
    * These are typed as RATE, so they can be safely summed over time
    * These aren't lossy—even a single request to a DB will be caught

### DIFF
--- a/debug.go
+++ b/debug.go
@@ -13,6 +13,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/DataDog/datadog-go/statsd"
 	"github.com/codahale/hdrhistogram"
 )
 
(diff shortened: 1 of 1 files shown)


dusty-stripe:
I feel like, if we haven't, we should check with #observability to see what they think about the count batching. If I understand correctly, datadog/veneur already sends metrics upstream on an interval instead of real time. And since the communication to the agent is local I wouldn't imagine it's a terribly large burden on Sequins. Otherwise, this LGTM.

vasi-stripe:
This is the approach OBS recommended when I asked them on Friday :)

PR #168, opened 2018-05-23
Open downloaded files for write with `O_DIRECT`.

bartle-stripe:
I hypothesize that some of our high system CPU usage can be attributed to caching effects.  Since we expect our working set to include a small subset of the just-downloaded data, it's counter-productive to keep the data in cache.

r? @vasi-stripe 
cc @stripe/storage


vasi-stripe:
lgtm

bartle-stripe:
PTAL @vasi I made this configurable, since 128 KiB seems better for NVMe-based AWS i3 instances.

PR #204, opened 2019-03-29
Update the zookeeper url to retrieve the package over TLS

nati-stripe:
Change the apache.org url in the travis config so that it pulls in zookeeper over HTTPS.


bartle-stripe:
lgtm

